PR #1, opened 2015-10-22
did not understand how to do this

LaurisaNeuwirth:


lib/assessment.rb
```diff
@@ -4,37 +4,38 @@ class Assessment
   # Write ruby code to return the longest entry in the
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
-  def longest(enum)
-    # Your code goes here
+  def longest(enum) { |a, b| a.length <=> b.length }
+    enum.max
```
GA-MEB, line 8: Close, actually! The block needs to be given as an argument to the enumerator's `max` method.

lib/assessment.rb
```diff
@@ -4,37 +4,38 @@ class Assessment
...
   end
 
+
   # Question 2
   # Write ruby code to return the sum of entries in the
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.length
```
GA-MEB, line 17: When summing across the elements in an array (or other enumerable), the tool to use is reduce. 
`enum.reduce(&:+)`

lib/assessment.rb
```diff
@@ -4,37 +4,38 @@ class Assessment
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.some?.odd?
```
GA-MEB, line 24: Close! `some` is JavaScript - the analogue in Ruby is `any?`. Also, the syntax to pass `odd?` into it is a little off.
`numbers.any?(&:odd?)`

lib/assessment.rb
```diff
@@ -4,37 +4,38 @@ class Assessment
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all?.even?
```
GA-MEB, line 31: Right idea! The syntax is a little off, though - do you think you can figure out how to fix it, based on the question above?

lib/assessment.rb
```diff
@@ -4,37 +4,38 @@ class Assessment
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    capitalized_entries = words.capitalize
```
GA-MEB, line 38: When you want to transform each element in a collection per some given rule, the tool to use is `map`. 
`words.map(&:capitalize)`

GA-MEB:
2 of 6 (-0.5 on Q1, Q3, and Q4; -1 on Q2 and Q6)

Not bad, but the syntax really tripped you up on this one. See if you can give it another go!

PR #7, opened 2015-10-22
completed assessment

mkutter72:


lib/assessment.rb
```diff
@@ -6,6 +6,7 @@ class Assessment
   # method that returns a Fixnum.
   def longest(enum)
     # Your code goes here
+    enum.max { |a,b| a.length <=> b.length}
```
GA-MEB, line 9: Correct.

lib/assessment.rb
```diff
@@ -14,6 +15,7 @@ def longest(enum)
   # Assume each entry implements a `+` method
   def sum(enum)
     # Your code goes here
+    enum.reduce(0) { |memo, item| memo + item}
```
GA-MEB, line 18: Correct. This could have been written even more shortly if you used the `&` trick:
`enum.reduce(&:+)`

lib/assessment.rb
```diff
@@ -21,13 +23,14 @@ def sum(enum)
   # in the numbers enumerable are odd
   def some_odd(numbers)
     # Your code goes here
+    numbers.reduce(false) {|memo, item| memo || item.odd?}
```
GA-MEB, line 26: Nice! I like seeing clever uses of `reduce`.
Best practice would probably be to do something like 
`numbers.any? {|num| num.odd?}`
or
`numbers.any?(&:odd?)`

lib/assessment.rb
```diff
@@ -21,13 +23,14 @@ def sum(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.reduce(true) {|memo, item| memo && item.even?}
```
GA-MEB, line 33: See above. Correct answer, but best practice would probably be something like 
`numbers.all? {|num| num.even?}`
or
`numbers.all?(&:even?)`

lib/assessment.rb
```diff
@@ -35,6 +38,7 @@ def every_even(numbers)
   # entries from the words enumerable capitalized.
   def transform(words)
     # Your code goes here
+    words.each {|x| x.capitalized}
```
GA-MEB, line 41: Yep! It could be even shorter if you used the `&` trick.
`words.map(&:capitalize)`

lib/assessment.rb
```diff
@@ -42,7 +46,10 @@ def transform(words)
   # named filename. Concatenate the lines together after
   # replacing the trailing newline character with a
   # space and return the result.
-  def read_file(filename)
-    # Your code goes here
-  end
+def read_file(filename)
+    File.open(ARGV[0]) do |file|
+
+      file.reduce('') {|memo, item| memo + item.chop + ' '}
```
GA-MEB, line 52: Mostly correct, but don't forget to remove the trailing space! One tool for that is `strip`.

GA-MEB:
5.75 of 6 (-0.25 on Q6)

Nice work, Mike!

lib/assessment.rb
GA-MEB, line 18: [EDIT] Actually, as of Ruby 2.0, it can be even shorter: just `enum.reduce(:+)`

lib/assessment.rb
GA-MEB, line 41: [EDIT] And in Ruby 2.0 and above, you can even leave off the `&`

PR #10, opened 2015-10-22
"completed assessment to the best of my abilities"

jcmeyer10:


jcmeyer10:
Curious as to how it came out with linter.  please advise.

lib/assessment.rb
```diff
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max_by { |x| x.length }.to_i
```
GA-MEB, line 8: Not the expected answer, but I think this will work. Why the `to_i`, though?

lib/assessment.rb
```diff
@@ -13,28 +13,30 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(&:+)
```
GA-MEB, line 16: Correct.

lib/assessment.rb
```diff
@@ -13,28 +13,30 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |x| x.odd? }
```
GA-MEB, line 23: Correct, minus the extra true underneath - that will cause the method to always return true.

lib/assessment.rb
```diff
@@ -13,28 +13,30 @@ def longest(enum)
...
+    true
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |x| x.even? }
```
GA-MEB, line 31: Correct, with the same caveat as above.

lib/assessment.rb
```diff
@@ -13,28 +13,30 @@ def longest(enum)
...
+    true
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    words.capitalized { |words| }
```
GA-MEB, line 39: Not quite. You want to perform a transformation across all the elements in `words` - what do you think the right tool would be for that?

GA-MEB:
3.25 of 6 (-0.25 on Q1, Q3, Q4; -1 on Q5, Q6)

Not bad, but I think you should give this one another shot. Sit on it for a day or so and try again!

lib/assessment.rb
jcmeyer10, line 8: I figured the to_i which as I understand it is to integer would return a
number since we wanted the number of the longest string not the string
itself.  Could have misunderstood.

lib/assessment.rb
jcmeyer10, line 39: Cramsters, things get squirrelly without Ruby Docs on hand.  I need
something that looks at every instance of the array and I would probably
put .capitalize after the block?

jcmeyer10:
Yah, I will.  Are these comments on my pull request?  I have a hard time
putting it all together when its so chopped up if that makes sense.

lib/assessment.rb
GA-MEB, line 39: Precisely! Something that iterates over every element, applies a transformative function/method, and gives us back an the collected results. Sound like anything you know?

lib/assessment.rb
GA-MEB, line 8: No, I think the question is asking for the longest thing, given whatever 'long' means in the context of that thing. Might be strings, for instance.

GA-MEB:
Yep, I can see them all there!

jcmeyer10:
Resubmitted today.  1-4 makes sense but would appreciate a five minute walk
through on five.  I put my thoughts in comments under the answer from
antony so you understood my level of understanding.

PR #11, opened 2015-10-22
DONE

julie1013:


lib/assessment.rb
```diff
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    entries.max { |a, b| a.length <=> b.length }
```

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    sum+=num
```
GA-MEB, line 16: No, sorry. When you want to sum across the elements in an array (or similar), which tool do you think is the one to use?

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    return numbers.some?(&:odd)
```
GA-MEB, line 23: Almost correct. `some` is JavaScript; `any?` is the analogue in Ruby.

Also, the return isn't actually necessary.

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    return numbers.some?(&:even)
```
GA-MEB, line 30: Should be `all?`.

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    words.upcase.split(' ').map { |word| word.gsub(/\W/, '') }
```
GA-MEB, line 37: You're right that you should use `map` here, but you also probably want to use a different method to capitalize each word. Can you think of how you might do that?

GA-MEB:
2.5 of 6 (-1 on Q2, -0.25 on Q3, -0.5 on Q4, -0.75 on Q5, -1 on Q6)

Take another look at this material - you're close on a number of these, and I think you can get there with a little more practice. Let me know if you'd like to resubmit!

PR #12, opened 2015-10-22
Incomplete response to question 6

kaylal12:


lib/assessment.rb
```diff
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max_by { |word| word.length }
```
GA-MEB, line 8: Correct! We also would have accepted 

lib/assessment.rb
```diff
@@ -13,28 +13,30 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce { |sum, num| sum += num }
```
GA-MEB, line 16: Correct. This could also have been written even more shortly if you used the `&` trick:
`enum.reduce(&:+)`

lib/assessment.rb
```diff
@@ -13,28 +13,30 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |num| num % 2 == 1 }
```
GA-MEB, line 23: Correct, but Ruby has a handy method called `.odd?` that could have made your life a bit easier.

lib/assessment.rb
```diff
@@ -13,28 +13,30 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |num| num % 2 == 0 }
```
GA-MEB, line 30: Correct, but see my note above re: `.odd?`/`.even?`

lib/assessment.rb
```diff
@@ -13,28 +13,30 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    words.map { |word| word.capitalize }
```
GA-MEB, line 37: Correct! Also could be
`words.map(&:capitalize)`

lib/assessment.rb
```diff
@@ -43,6 +45,6 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    filename.each_with_object { |word| word.replace("\n", ' ') }
```
GA-MEB, line 48: Mostly correct; just don't forget to trim the extra whitespace at the end!

GA-MEB:
5.75 of 6 (-0.25 on Q6)

Nice work.

lib/assessment.rb
GA-MEB, line 16: [EDIT] Actually, as of Ruby 2.0, it can be even shorter: just `enum.reduce(:+)`

lib/assessment.rb
GA-MEB, line 37: [EDIT] If using Ruby 2.0 or above, you could leave off the &.

PR #14, opened 2015-10-22
responses to the ruby enumerables assessment

christinemoses:


lib/assessment.rb
```diff
@@ -5,7 +5,8 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max_by { |x| x.length }
+    # I didn't figure out how to solve this, but was testing ways to use the above line found at http://ruby-doc.org/core-2.1.0/Enumerable.html#method-i-max_by
```
GA-MEB, line 9: Researching is almost as good as knowing; even though you didn't fully understand, great job digging!

lib/assessment.rb
```diff
@@ -13,14 +14,18 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce {0, :+}
```
GA-MEB, line 17: You're right that `.reduce` is the right tool, and your syntax is _almost_ right.
`enum.reduce(0,&:+)` (or `enum.reduce(0,:+)`, as of Ruby 2.0)

lib/assessment.rb
```diff
@@ -13,14 +14,18 @@ def longest(enum)
...
+    # and the enumerable exercise we did with fibonacci in class yesterday
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.select(&:odd?)
+    # I based this answer off of the enumerables exercise we did in class yesterday with fibonacci numbers to find the sum of odd numbers: fibs.select(&:odd?).reduce(&:+), "\n"
```
GA-MEB, line 28: Good thinking, and your answer is close, but `select` is not the right tool. Can you think of one that might be a better fit?

GA-MEB:
2.25 of 6

Definitely room for improvement here, but the stuff that you submitted wasn't far off. Keep at it, and let me know if you plan to resubmit this one!

PR #15, opened 2015-10-22
completed assessment

sarabrentmccoy:


lib/assessment.rb
```diff
@@ -5,36 +5,52 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max {|a, b| a.length <=> b.length}
+    end
```
GA-MEB, line 9: Correct (ignoring the extra `end`)

lib/assessment.rb
```diff
@@ -5,36 +5,52 @@ class Assessment
...
 
+  puts longest
+
   # Question 2
   # Write ruby code to return the sum of entries in the
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce{|a, b| a + b}
```
GA-MEB, line 19: Correct (again ignoring the extra `end`)
Another way to write it would have been to use the `&` shortcut:
`enum.reduce(&:+)`
In fact, in Ruby 2.0 and above, you can even leave off the `&`:
`enum.reduce(:+)`

lib/assessment.rb
```diff
@@ -5,36 +5,52 @@ class Assessment
...
   end
 
+  puts sum
+
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? {|number| number%2 == 1}
```
GA-MEB, line 29: Correct! You could also have used `.odd?`

GA-MEB, line 29: That would actually allow you to write 
`numbers.any?(&:odd?)` (or just `numbers.any?(:odd?)` in Ruby 2.0 and above)

lib/assessment.rb
```diff
@@ -5,36 +5,52 @@ class Assessment
...
   end
 
+  puts some_odd
+
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? {|number| number%2 == 0}
```
GA-MEB, line 39: Correct. See above, except with `.even?`

lib/assessment.rb
```diff
@@ -5,36 +5,52 @@ class Assessment
...
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
+
+  #help
   def transform(words)
-    # Your code goes here
+    memo = 0
+    upcase.map {|memo, word| memo + word
+    end
```
GA-MEB, line 53: You're thinking along the right lines, but you've put them together incorrectly. When you use `.map`, what gets transformed? What determines how the transformation works?

Also, are you sure `.upcase` is the right tool to use if we just want the words to be capitalized (as opposed to being in all-caps)?

lib/assessment.rb
```diff
@@ -43,6 +59,13 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    #I don't really understand the ARGV[0], just took from read-file.rb demo yesterday, and I don't know when to use it/if I need it here.
```
GA-MEB, line 62: ARGV is a list of arguments passed in when a program is run from the console. You'll already have the filename you need (passed in as an argument to the method) so reading from the console is unnecessary.

lib/assessment.rb
```diff
@@ -43,6 +59,13 @@ def transform(words)
...
+    File.open(ARGV[0]) do |filename|
```
MattB-GA-Boston: When you use File.open, what should you pass in as an argument?  And when it does its iteration, what are the specific things that it's iterating through?

lib/assessment.rb
```diff
@@ -43,6 +59,13 @@ def transform(words)
...
+      filename.reduce(&:+)
+      filename.gsub(/\n/, ' ')
+    end
```
MattB-GA-Boston: You've got a lot of the pieces here, they're just not together in the right order.

GA-MEB:
4.75 (-0.75 on Q5, -0.5 on Q6)

Not bad! Take another look at Q5 and Q6, and see if you can figure them out.

lib/assessment.rb
GA-MEB, line 66: Please disregard my previous notes on this question - I was mistaken about something with `File.open`. As written, it's still not quite right (replacing newlines with spaces should happen before concatenation, and you still need to strip off whitespace at the end); however, I'm giving you half-credit on this one now.

PR #16, opened 2015-10-22
Finished the assessment...well not really

AkJones2007:


lib/assessment.rb
```diff
@@ -14,6 +16,7 @@ def longest(enum)
   # Assume each entry implements a `+` method
   def sum(enum)
     # Your code goes here
+    enum.reduce(&:+)
```
GA-MEB, line 19: Correct.

GA-MEB:
No kidding. For what it's worth, you did get the question you answered totally correct. Give this one another go - I think you'll do better now that you've had some time to digest it.

PR #17, opened 2015-10-22
Finished assessment

LaurenGrant:


lib/assessment.rb
```diff
@@ -5,37 +5,42 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max { |a, b| a.length <=> b.length }
```

lib/assessment.rb
```diff
@@ -5,37 +5,42 @@ class Assessment
...
   # Question 2
   # Write ruby code to return the sum of entries in the
-  # enumerable supplied (i.e. apply the `+` operator)
+  # enumerable supplied
+  #(i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+# What enumerable supplied? Is there something we should actually be working off of?
+  enum.sum #or use the inject method?
```
GA-MEB, line 18: `inject` (or, equivalently, `reduce) is the right tool to use here, not`sum`. Can you think of how you'd set it up?

lib/assessment.rb
```diff
@@ -5,37 +5,42 @@ class Assessment
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    !numbers.any? &block
```
GA-MEB, line 25: `.any?` is the right tool here, but I'm confused why you're negating the whole expression. Also, where is `block` defined?

lib/assessment.rb
```diff
@@ -5,37 +5,42 @@ class Assessment
...
   end
+  print numbers
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? &block
```
GA-MEB, line 33: See comment above about `block`. Otherwise, your setup is correct.

lib/assessment.rb
```diff
@@ -5,37 +5,42 @@ class Assessment
...
   end
+  print numbers
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    words.capitalized
```
GA-MEB, line 41: You need to iterate through all of the words and apply a transformation to them (specifically, capitalization). What enumerable methods do you think you might use to accomplish that?

lib/assessment.rb
```diff
@@ -43,6 +48,8 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    file.open("filename.txt") {|f| f.readline}
```
MattB-GA-Boston: When you open a file using `File` the file is read as an collection of ... what? What is each element?

lib/assessment.rb
```diff
@@ -43,6 +48,8 @@ def transform(words)
...
+    filename.gsub(/\n/, ' ') # replaces newline with a space
```
MattB-GA-Boston: `filename` is just text: a path to the file that's being opened.

GA-MEB:
2 of 6 (-1 on Q2, Q5, Q6; -0.5 on Q3, Q4)

Keep at it - you're not far off on a lot of these. Let me know if/when you'd like to try this one again.

lib/assessment.rb
```diff
@@ -5,37 +5,42 @@ class Assessment
...
 
   # Question 2
   # Write ruby code to return the sum of entries in the
-  # enumerable supplied (i.e. apply the `+` operator)
+  # enumerable supplied
+  #(i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+# What enumerable supplied? Is there something we should actually be working off of?
```
GA-MEB, line 17: `enum` is the enumerable supplied.

lib/assessment.rb
GA-MEB, line 51: Oops, my mistake - disregard my previous comment.

GA-MEB, line 51: However, this will still not work - the code for handling the file's contents must be found inside the block _only_. Additionally, `readline` does not look like it's a valid method for a File to call.

PR #18, opened 2015-10-22
finished assessment

raq929:


lib/assessment.rb
```diff
@@ -5,7 +5,9 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.reduce(0) do |memo, item|
+      item.length > memo ? item.length : memo
+    end
```
GA-MEB, line 10: This approach is _almost_ right - the problem is that it will return the _length_ of the longest element, rather than the element itself. Can you think of how you might change your code to get that result?

lib/assessment.rb
```diff
@@ -13,28 +15,30 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(0) do |memo,item|
```
GA-MEB, line 18: Yep, that's correct. You could also have written it as 
`enum.reduce(0, &:+)` (or just `enum.reduce(0, :+)`, as of Ruby 2.0)

lib/assessment.rb
```diff
@@ -13,28 +15,30 @@ def longest(enum)
...
+      memo + item
+    end
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? {|number| number.odd? }
```
GA-MEB, line 27: Correct. The shorter version would be
`numbers.any?(&:odd?)` / `numbers.any?(:odd?)`

lib/assessment.rb
```diff
@@ -13,28 +15,30 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |number| number.even? }
```
GA-MEB, line 34: Correct. See above for the format of an equivalent shorthand.

lib/assessment.rb
```diff
@@ -13,28 +15,30 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    words.map {|word| word.upcase}
```
GA-MEB, line 41: Ooh, so close! We wanted the words to be capitalized ("Apple") not all caps ("APPLE"). Can you think of a tool that would do that for you?

lib/assessment.rb
```diff
@@ -43,6 +47,8 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    file.open(filename) do |file|
+      file.reduce(&:+).gsub(/"\n"/, " ")
```
GA-MEB, line 51: Your order is backwards - you were supposed to do the gsub before concatenating. Also, don't forget to remove whitespace at the end.

GA-MEB:
4.5 of 6 (-0.5 on Q1, Q5, Q6)

Nice work!

PR #20, opened 2015-10-22
response

faetea:


lib/assessment.rb
```diff
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum(&:size).max.last
```
GA-MEB, line 8: Not quite right. `enum` is an Enumerable object, not a method, so it can't accept any arguments. Instead, you'll need to call the methods that are on it.

lib/assessment.rb
```diff
@@ -13,28 +13,29 @@ def longest(enum)
...
-    # Your code goes here
+    self%1==0 && self.to_i.even?
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    words.map(&:upcase)
```
GA-MEB, line 38: Close; `upcase` isn't the right tool.

lib/assessment.rb
```diff
@@ -13,28 +13,29 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(0) do |memo, enum|
+      memo + 1
```
GA-MEB, line 17: `reduce` is the right tool to use here, but your setup is wrong.

lib/assessment.rb
```diff
@@ -13,28 +13,29 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    self%1==0 && self.to_i.odd?
```
GA-MEB, line 24: This answer and the next one are both incorrect.

GA-MEB:
1 of 6 (1/2 credit on Q5, 1/4 credit on Q1 and Q2) 

You might want to review the material and take this one again; I'd be happy to give you feedback on your resubmission.

PR #21, opened 2015-10-22
Completed assessment.

atsundwn:


lib/assessment.rb
```diff
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.reduce { |a, b| a.length <=> b.length }
```
GA-MEB, line 8: Close. With this kind of block, you don't want to use `reduce`, but rather an alternative method. However, it is possible to solve this one with `reduce`.

lib/assessment.rb
```diff
@@ -13,28 +13,36 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(&:+)
```
GA-MEB, line 16: Correct. In fact, as of Ruby 2.0, you can even drop the `&` and just write
`enum.reduce(:+)`

lib/assessment.rb
```diff
@@ -13,28 +13,36 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    if numbers.each.odd?
```
GA-MEB, line 23: The Enumerator returned by `.each` does not have an `.odd?` method.

lib/assessment.rb
```diff
@@ -13,28 +13,36 @@ def longest(enum)
...
+      true
+    else
+      false
+    end
```
GA-MEB, line 27: There is another method you should use here - one which determines if at least some elements in a given set pass some test. Can you think of what it might be?

lib/assessment.rb
```diff
@@ -13,28 +13,36 @@ def longest(enum)
...
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    if numbers.each.even? == false
+      false
+    else
+      true
+    end
```
GA-MEB, line 38: Incorrect, for the same reasons as Q2.

lib/assessment.rb
```diff
@@ -13,28 +13,36 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    return words.map { |a| a.upcase }
```
GA-MEB, line 45: Not `upcase` but ... ?

GA-MEB, line 45: Also, the `return` is not necessary here; by default, Ruby will always return the last evaluated expression in your method.

lib/assessment.rb
```diff
@@ -43,6 +51,8 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    text = filename.each.gsub("\n" , " ").join(" ")
```
GA-MEB, line 54: `filename` is just a string, giving the path to a particular file. It is not the actual contents of the file.

GA-MEB:
2 of 6 (-0.5 on Q1, Q5; -1 on Q3, Q4, Q6)

It looks like you had some trouble with this one, Min. Give the material another read-through, and resubmit this assessment when you get a chance.

PR #22, opened 2015-10-22
Assessment Complete. Maybe?

kelvinma:
Lots o' Ruby Doc! 

And some class notes.

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max { |a, b| a.length <=> b.length }
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/assessment.rb
```diff
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max { |a, b| a.length <=> b.length }
```

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(&:+)
```
GA-MEB, line 16: Correct. In fact, as of Ruby 2.0, you can drop the `&` and just write
`enum.reduce(:+)`

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any?.&:odd?
```
GA-MEB, line 23: No `.`, but otherwise right.

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all?.&:even?
```
GA-MEB, line 30: Ditto.

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    words.split(' ').map(&:capitalize)
```
GA-MEB, line 37: `words` is an enumerable; it doesn't have a `split` method, and wouldn't need one anyway. Just `map` and `capitalize` is enough.

lib/assessment.rb
```diff
@@ -43,6 +43,6 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    filename.each_with_object([]) { |line| line }.chomp.join(' ')
```
GA-MEB, line 46: Good thinking, but not correct.

GA-MEB:
4 of 6 (-0.25 on Q3, Q4; -0.5 on Q5; -1 on Q6)

Nice work, Kelvin.

PR #23, opened 2015-10-22
Almost completed assessment.

tictacktone:


lib/assessment.rb
```diff
@@ -5,7 +5,8 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    do enum.max{|a, b| a.length <=> b.length}
```
GA-MEB, line 8: Almost. A `do` block (`do` ... `end`) can't be executed on its own in Ruby; it needs to be executed by a method, or else stored for future use by something called a Proc (more on this to come soon). However, if you remove the `do` and `end`, this code will in fact work correctly.

lib/assessment.rb
```diff
@@ -13,28 +14,29 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(:+)
```
GA-MEB, line 17: Correct; however, this will only work in Ruby 2.0 and above. For older versions of Ruby, you will need to explicitly call `&` before the `:+` symbol.

lib/assessment.rb
```diff
@@ -13,28 +14,29 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.odd?
```
GA-MEB, line 24: `odd?` is one of the tools you'll need, but there's a second. What enumerable method could be used to determine if at least some elements in a collection pass some particular test?

lib/assessment.rb
```diff
@@ -13,28 +14,29 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+   numbers.even?
```
GA-MEB, line 31: See above. `even?` is right. What enumerable method could be used to determine if _every_ element in a collection passes a particular test?

lib/assessment.rb
```diff
@@ -13,28 +14,29 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+     words.upcase
```
GA-MEB, line 38: Sorry, `upcase` is not the right tool - we want to capitalize ("Apple") not put everything in caps ("APPLE").

Additionally, there's another piece missing: an enumerable method to perform a transformation across all elements in a collection. Can you think of which it might be?

GA-MEB:
3 of 6 (-0.25 on Q1; -0.5 on Q3, Q4; -0.75 on Q5; -1 on Q6)

Not bad, TJ - you're pretty close on a couple of these answers. Take another look over this material and try again when you have a chance.

PR #24, opened 2015-10-22
Commiting my work

gracenjambi:
commiting work

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -5,7 +5,10 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    longest = %w{ cat sheep bear }.inject do |memo, word|
+    memo.length > word.length ? memo : word
+    end
+    longest
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/assessment.rb
```diff
@@ -5,7 +5,10 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    longest = %w{ cat sheep bear }.inject do |memo, word|
+    memo.length > word.length ? memo : word
+    end
+    longest
```
GA-MEB, line 11: Your approach is generally correct - however, we want to do operate on the collection `enum`, not just any collection that we might pick. 

Additionally, inject will already return the last result, so the final line there is unnecessary.

lib/assessment.rb
```diff
@@ -13,28 +16,31 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    (5..10).inject { |sum, n| sum + n }
```
GA-MEB, line 19: Again, you're correct in using  `+` and `inject` (or, equivalently, `reduce`), but we want to operate on enum.
`enum.inject { |sum, n| sum + n  }`

lib/assessment.rb
```diff
@@ -13,28 +16,31 @@ def longest(enum)
...
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
-  end
+    a = [7, 5, 9, 2, 0, 7, 9, 4, 2, 0]
+    p a.slice_when {|i, j| i.even? != j.even? }.to_a
+
```
GA-MEB, line 28: No, sorry - this is not correct.

First, we want to be operating on the parameter `numbers` that we've been given.

Second, we need to use an enumerable method to determine if at least some elements in our collection pass a particular test. Can you think of one that would behave in that way?

Third, when we actually do the test, we want to test whether or not any given element is odd, so `!x.even?` is ok, but there's another method we could use which would be even shorter...

lib/assessment.rb
```diff
@@ -13,28 +16,31 @@ def longest(enum)
...
+    end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
-  def every_even(numbers)
-    # Your code goes here
-  end
+    [1,2,3,4,5].select { |num|  num.even?  }
```
GA-MEB, line 34: Your test is correct, but your enumerable method is not. Also, where is the method definition? We want to be able to call `every_even` on any given set of numbers and get back either `true` or `false`

GA-MEB:
1.75 of 6 (-0.25 on Q1, Q2; -1 on Q3, Q5, Q6; -0.75 on Q4)

Take another look through the lesson, and try this one again when you have some time. And let me know when you resubmit, so that I can give you feedback!

PR #25, opened 2015-10-22
More or less Had answers for first 5 but really no indea on the last one

bbarbersox:
still very confused on enumerables

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max { |a, b| a.length <=> b.length }
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/assessment.rb
```diff
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max { |a, b| a.length <=> b.length }
```

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.inject { |a, b| a + b }
```
GA-MEB, line 16: Correct. You could also have written
`enum.inject(&:+)` (or, in Ruby 2.0 and above, just `enum.inject(:+)`)
And of course, `reduce` would have worked exactly the same.

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    enum.each.odd?
```
GA-MEB, line 23: Sorry, that's not right. First, there is a method to see if at least some members of a collection pass a particular test. Second, The syntax is incorrect.

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    enum.every.even?
```
GA-MEB, line 30: Right methods this time, just wrong syntax!

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    words_array = words.map(&:capitalize)
```
GA-MEB, line 37: The variable assignment is unnecessary, but this will give the right answer.

GA-MEB:
3.75 of 6 (-0.72 on Q3, 0.5 on Q4, -1 on Q6)

Not bad, but I agree that there's room for improvement. Take another look over the lesson and see if it makes more sense now that you've had some time to digest it.

PR #26, opened 2015-10-22
completed assessment

cwcrane:


lib/assessment.rb
```diff
@@ -5,7 +5,12 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.reduce do |item, memo|
```
GA-MEB, line 8: These are in the wrong order. According to the documentation for `reduce`, its signature is `reduce { |memo, item| block } → item`

lib/assessment.rb
```diff
@@ -5,7 +5,12 @@ class Assessment
...
+      if memo.to_s.length > item.to_s.length
```
GA-MEB, line 9: We don't know that these aren't strings already. All we know is that whatever kind of data is found in `enum`, it has a `.length` method.

lib/assessment.rb
```diff
@@ -5,7 +5,12 @@ class Assessment
...
+        longest = memo
+      else longest = item
+      end
+    end
```
GA-MEB, line 13: Otherwise, this is correct. However, it could have been written much more shortly. Let's do some refactoring - we can replace the if..else with a ternary operation and condense the whole thing to a single line.
`enum.reduce { |memo, item| memo.length > item.length? memo : item }`

lib/assessment.rb
```diff
@@ -13,28 +18,29 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce do |item, memo|
+      memo += item
```
GA-MEB, line 22: `reduce` and `+` are the right tools. However, `+=` is not necessary, because the value that gets put into `memo` the next time around is what the block _returns_, and both `memo + item` and `memo += item` will return the same value.

GA-MEB, line 22: Missing an `end` as well.

lib/assessment.rb
```diff
@@ -13,28 +18,29 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any?{|x| x.odd?}
```
GA-MEB, line 29: Correct! This could also have been written as 
`numbers.any?(&:odd?)` 
(or, in Ruby 2.0 and above, as just `numbers.any?(:odd?)`)

lib/assessment.rb
```diff
@@ -13,28 +18,29 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all?{|x| x.even?}
```
GA-MEB, line 36: Correct. See my note above re: alternate syntax.

lib/assessment.rb
```diff
@@ -13,28 +18,29 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    words.map{|x| x.upcase)
```
GA-MEB, line 43: Not `upcase` - that will take "apple" and turn it into "APPLE". What we want is to capitalize it, turning it into "Apple".

lib/assessment.rb
```diff
@@ -43,6 +49,7 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    File.open(filename) do |file|
+      file.gsub(/"\n"/, ' ').each_with_object([]){|i, a| i << a}
```
GA-MEB, line 53: Very close! You should also trim that final space at the end of the last line - the method to do that is `strip` (or `rstrip`, since it's specfically trailing whitespace that we're talking about).

GA-MEB:
4.25 of 6 (-0.5 on Q1, Q2, Q5; -0.25 on Q6)

Nice work, Chris.

PR #27, opened 2015-10-22
Finish

dhuddell:
Finish assessment

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -5,7 +5,9 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    long = 0
+    enum.reduce(0) { |prev, current| current > prev ? long = current : long}
+    puts long
   end
 
   # Question 2
@@ -13,28 +15,28 @@ class Assessment
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(:+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? {|number| number.odd? }
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? {|number| number.even? }
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    words.to_a {|word| word.capitalize }
   end
 
   # Question 6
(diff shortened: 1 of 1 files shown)


lib/assessment.rb
```diff
@@ -5,7 +5,9 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    long = 0
+    enum.reduce(0) { |prev, current| current > prev ? long = current : long}
+    puts long
```
GA-MEB, line 10: You're thinking along the right lines, but there are a couple of problems with this answer. 
1. You're comparing `current` and `prev` directly, instead of comparing their lengths.
2. This method will return `nil` instead of the longest element.
3. The `long` variable is ultimately unnecessary, since `reduce` will automatically return the last result of its block.

lib/assessment.rb
```diff
@@ -13,28 +15,28 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(:+)
```
GA-MEB, line 18: Correct, though this will only work in Ruby 2.0 and above; in older versions of Ruby, you must use `&` before the symbol:
`enum.reduce(&:+)`

lib/assessment.rb
```diff
@@ -13,28 +15,28 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? {|number| number.odd? }
```
GA-MEB, line 25: Correct. You could also use the shorthand notation that you used in Q2:
`numbers.any?(:odd?)`

lib/assessment.rb
```diff
@@ -13,28 +15,28 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? {|number| number.even? }
```
GA-MEB, line 32: Correct.

lib/assessment.rb
```diff
@@ -13,28 +15,28 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    words.to_a {|word| word.capitalize }
```
GA-MEB, line 39: `capitalize` is right, but the first part isn't. What enumerable method would we use to perform a transformation (in this case, capitalization) on each element in a collection?

lib/assessment.rb
```diff
@@ -43,6 +45,6 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+   File.open(filename).map(&:chomp).join(" ")
```
GA-MEB, line 48: Hah! Clever - since the newline character will always be last in the line, `chomp` will get rid of every one, and using will ensure that there's no leftover whitespace at the end (unless, of course, there's trailing whitespace in the last line.... but let's ignore that case for now). Well done.

GA-MEB:
5 of 6 (-0.5 on Q1, Q5)

Nice work!

PR #28, opened 2015-10-22
Answered q's

HoldenMills:


lib/assessment.rb
```diff
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max { |memo, word| memo.length <=> word.length }
```

lib/assessment.rb
```diff
@@ -13,21 +13,21 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(:+)
```
GA-MEB, line 16: Correct, but only in Ruby v2.0 and above. In older versions of Ruby, you need to explicitly add the `&` before the symbol:
`enum.reduce(&:+)`

lib/assessment.rb
```diff
@@ -13,21 +13,21 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |numbers| (numbers % 2) != 0 }
```
GA-MEB, line 23: Correct. You could also use the `.odd?` method, e.g.
`numbers.any? {|num| num.odd?}`
or even
`numbers.any?(:odd?)`

lib/assessment.rb
```diff
@@ -13,21 +13,21 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all { |numbers| (numbers % 2) == 0 }
```
GA-MEB, line 30: Correct. See my previous note about alternative ways to write it.

GA-MEB:
4 of 6 (-1 on Q5, Q6)

Nice work, Holden!

PR #29, opened 2015-10-22
assessment

drewster317:


lib/assessment.rb
```diff
@@ -5,7 +5,10 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+
+    ar = ['one','two','three','four','five']
+    ar.max_by(&:length)
+
```
GA-MEB, line 11: Nice! The only problem is that you were supposed to do it to `enum` instead of your own array.

lib/assessment.rb
```diff
@@ -13,28 +16,37 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+
+    array.each { |a| sum += a }
```
GA-MEB, line 20: Ditto w.r.t. `enum`. 

Technically this will work, but the `+=` is unnecessary; `+` would work just fine.

lib/assessment.rb
```diff
@@ -13,28 +16,37 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+
+    array = []
+    array.map!(&:capitalized)
```
GA-MEB, line 48: Should be on `words` instead of your own array, and the method is called `capitalize`, not `capitalized`

lib/assessment.rb
```diff
@@ -13,28 +16,37 @@ def longest(enum)
...
+
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+
+    self.values_at(* self.each_index.select {|i| i.odd?})
```
GA-MEB, line 29: `self` is like `this` in JS - it's used to refer back to objects from methods/functions. Since this method is top-level, `self` will refer to the top-level object, `main`, and sadly, `main` does not have a `values_at` method. 

> However, if you were, say, reopening the Array class and writing a new method on it, Array _does_ have a method called `values_at` that you could call. Did you look this up on Stack Overflow by any chance?

lib/assessment.rb
```diff
@@ -13,28 +16,37 @@ def longest(enum)
...
+
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+
+    self.values_at(* self.each_index.select {|i| i.even?})
```
GA-MEB, line 38: See above.

GA-MEB:
2 of 6 (-0.25 on Q1, Q2; -1 on Q3, Q4, Q6; -0.5 on Q5)

Your first two were pretty close - it was Q3 and Q4 that really knocked you for a loop. Take a look back over the material and give this one another shot when you have time.

PR #30, opened 2015-10-22
Response

francescadal:
I had difficulty with the syntax and transferring what i thought into code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,6 @@ Follow the prompts in `lib/assessment.rb` and complete each question.  You may u
 
 **Unless explicilty disallowed, you may use any method or feature ruby provides to complete the prompts.**
 
-Check your answers by running `grunt test`.  Check your code by running `grunt nag`. You can reformat your code with `grunt reformat`.
+Check your answers by running `rspec`.
 
 When you are done, commit, push your changes, and submit a pull request.
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+     enum.max { |a, b| a.length <=> b.length}
   end
 
   # Question 2
@@ -20,21 +20,21 @@ class Assessment
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    odd?
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    even?
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+   words.length.upcase
   end
 
   # Question 6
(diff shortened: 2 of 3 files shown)


lib/assessment.rb
```diff
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+     enum.max { |a, b| a.length <=> b.length}
```

lib/assessment.rb
```diff
@@ -20,21 +20,21 @@ def sum(enum)
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    odd?
```
GA-MEB, line 23: Yes, you will want to use `odd?`, but you want to use that method to evaluate each element in `enum`. Are there any enumerable methods that we might use to check if at least some elements in our collection pass a given test?

lib/assessment.rb
```diff
@@ -20,21 +20,21 @@ def sum(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    even?
```
GA-MEB, line 30: As above, `even?` is one of the right tools for the job, but you're missing a key piece. Which enumerable method might we use to check if _every_ element passes a given test?

lib/assessment.rb
```diff
@@ -20,21 +20,21 @@ def sum(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+   words.length.upcase
```
GA-MEB, line 37: No, this isn't right. `words.length` will give us a number, the length of `words`, and trying to capitalize a number doesn't make sense.

What we want to do in this question is iterate through `words` and apply a transformation (specifically, capitalization) to each word. Can you think of what tools we might need in order to be able to do that?

README.md
```diff
@@ -12,6 +12,6 @@ Follow the prompts in `lib/assessment.rb` and complete each question.  You may u
 
 **Unless explicilty disallowed, you may use any method or feature ruby provides to complete the prompts.**
 
-Check your answers by running `grunt test`.  Check your code by running `grunt nag`. You can reformat your code with `grunt reformat`.
+Check your answers by running `rspec`.
```
MattB-GA-Boston: Nice - did you end up running the tests? Were they helpful?

spec/assessment_spec.rb
```diff
@@ -47,7 +47,7 @@
       expect(@response.some_odd([2, 3, 4, 6])).to eq(true)
     end
     it 'returns true when passed 2..6' do
-      expect(@response.every_even(2..6)).to eq(true)
+      expect(@response.some_odd(2..6)).to eq(true)
```
MattB-GA-Boston: Huh. Were these errors in the tests? If so, great job catching them! Can you please add an issue to this assessment with your changes to the tests?

GA-MEB:
1.5 of 6 (-0.75 on Q3, Q4, Q5; -1 on Q2, Q6)

Yeah, it looks like this one knocked you for a loop. Take another read through the lesson and try this assessment again when you have some time. Let me know if you have any questions about my feedback.

PR #31, opened 2015-10-22
added responses

resource11:


lib/assessment.rb
```diff
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    max { |a, b| a.length <=> b.length }
```
GA-MEB, line 8: Close - `enum.max`

lib/assessment.rb
```diff
@@ -13,28 +13,32 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    reduce (memo, f) { return memo + f }
```
GA-MEB, line 16: `enum.reduce`. Also, Ruby will implicitly return the last value in a block, so the `return` there is unnecessary.

lib/assessment.rb
```diff
@@ -13,28 +13,32 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
+
+  # Yeah... confused
   def some_odd(numbers)
-    # Your code goes here
+    each_with_index { |number, index| number.odd? }
```
GA-MEB, line 25: `odd?` is the right tool, but you used the wrong enumerable method. Which one would we use if we wanted to determine if at least some elements in our collection passed a particular test?

lib/assessment.rb
```diff
@@ -13,28 +13,32 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
+
+  # same comment as above
   def every_even(numbers)
-    # Your code goes here
+    each_with_index { |number, index|  number.even? }
```
GA-MEB, line 34: See above; `even?` is correct, but your enumerable method is wrong. Which method tells us if _every_ element in a particular set passes some test?

lib/assessment.rb
```diff
@@ -13,28 +13,32 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    upcase
```
GA-MEB, line 41: If we want to perform a transformation on every element in a collection, which enumerable method would we want to use?

As far as the transformation itself, `upcase` will turn "apple" into "APPLE", but what we want is "Apple". Are there any string methods we could use to do that?

GA-MEB:
2 of 6 (-0.25 on Q1, Q2; -0.75 on Q3, Q4; -1 on Q5, Q6)

Per your self-assessment on Friday, I think you can definitely do some more work here, but I also don't think that you're as far off as you think you are. Now that you've had a few days to digest on the lesson and reflect, do you think you'd be ready to try this one again?

lib/assessment.rb
resource11, line 8: Oh! We need to append the passed in argument with a method. New ruby syntax goodness to learn.

lib/assessment.rb
resource11, line 16: The implicit return. Love that!

PR #33, opened 2015-10-24
Assessment

jcmeyer10:
"amended second and third answers so they wouldn't always return true …
and corrected answer 4 and annotated my thoughts on answer 5 to help
clarify to instructors my understanding of material"

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -21,7 +21,6 @@ class Assessment
   # in the numbers enumerable are odd
   def some_odd(numbers)
     numbers.any? { |x| x.odd? }
-    true
   end
 
   # Question 4
@@ -29,15 +28,17 @@ class Assessment
   # numbers enumerable are even
   def every_even(numbers)
     numbers.all? { |x| x.even? }
-    true
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    words.capitalized { |words| }
+    words.map(&:capitalize)
   end
+
+  # the reason why we can use the &:capitalize after words.map is
+  # because it operates the same was as words.map { |w| w.capitalize }
 
   # Question 6
   # Write ruby code to read each line from the file
@@ -45,6 +46,14 @@ class Assessment
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    File.open(filename).reduce(' ') { |a, e| a << e.chomp + ' ' }.strip
   end
 end
+
+# admittedly I needed some help with this one but here is my understanding:
+# We use File.open and then pass it the filename from the perameters
+# then use reduce to combine all the elements of the file and then use reduce
+# and pass it an initial value of '' and pass the perameters a and e
+# we then use the << to append the argument we get from e.chomp and adding
+# an empty string to seperate the words
+# we end by using strip to remove all trailing and leading whitespace
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max_by { |x| x.length }.to_i
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/assessment.rb
```diff
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max_by { |x| x.length }.to_i
```
GA-MEB, line 8: We may be dealing with strings here; enum could be `["a", "ab", "abc"]`. In that case, performing a `.to_i` on the longest string doesn't make sense.

Otherwise, this is reasonable.

lib/assessment.rb
```diff
@@ -13,36 +13,47 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(&:+)
```
GA-MEB, line 16: Correct.

lib/assessment.rb
```diff
@@ -13,36 +13,47 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |x| x.odd? }
```
GA-MEB, line 23: Correct. Alternatively,
`enum.any?(&:odd?)` 
(or, for Ruby 2.0 and above, `enum.any?(:odd?)`)

lib/assessment.rb
```diff
@@ -13,36 +13,47 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |x| x.even? }
```
GA-MEB, line 30: Correct; see my note above about alternatives.

lib/assessment.rb
```diff
@@ -13,36 +13,47 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    words.map(&:capitalize)
```
GA-MEB, line 37: Correct.

lib/assessment.rb
```diff
@@ -13,36 +13,47 @@ def longest(enum)
...
+  # because it operates the same was as words.map { |w| w.capitalize }
+
   # Question 6
   # Write ruby code to read each line from the file
   # named filename. Concatenate the lines together after
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    File.open(filename).reduce(' ') { |a, e| a << e.chomp + ' ' }.strip
```
GA-MEB, line 49: We probably want to start with an empty string here, not a space. Nevertheless, this is a valid solution; in fact `strip` will also take out the leading whitespace.

GA-MEB, line 49: You could also just use `+` instead of `<<`, and that might be more idiomatic. You could even do the whole thing using string interpolation:
`"#{a} #{e.chomp}"` 
or 
`"#{a}#{e.chomp} "`
would both work, as long as you're running `strip` at the end to take off any extra leading or trailing whitespace.

lib/assessment.rb
```diff
@@ -13,36 +13,47 @@ def longest(enum)
...
   end
 end
+
+# admittedly I needed some help with this one but here is my understanding:
+# We use File.open and then pass it the filename from the perameters
+# then use reduce to combine all the elements of the file and then use reduce
+# and pass it an initial value of '' and pass the perameters a and e
+# we then use the << to append the argument we get from e.chomp and adding
+# an empty string to seperate the words
+# we end by using strip to remove all trailing and leading whitespace
```
GA-MEB, line 59: Yep! This is a great step-by-step explanation of what's going on. Nice job.

GA-MEB:
5.75 of 6 (-0.25 on Q1)

Excellent work, Jeff! You're making great progress.

PR #34, opened 2015-10-25
Re-submitting assessment

atsundwn:


lib/assessment.rb
```diff
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max { |a, b| a.length <=> b.length }
```

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(:+)
```
GA-MEB, line 16: Correct, for Ruby 2.0 and above; for older versions of Ruby, you need to include an `&` before the symbol.
`enum.reduce(&:+)`

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |i| i.odd? }
```
GA-MEB, line 23: Correct. You can also write
`numbers.any? (&:odd?)`

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |i| i.even? }
```
GA-MEB, line 30: Correct. See above for a shorter version.

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    words.map { |a| a.capitalize }
```
GA-MEB, line 37: Correct.

lib/assessment.rb
```diff
@@ -43,6 +43,6 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    File.open(filename).reduce('') { |memo, line| memo + line.chomp + ' ' }.strip
```
GA-MEB, line 46: This will mostly work! There's just one problem - what if the last line doesn't end in a newline character? You may need to account for that case in your solution.

GA-MEB, line 46: Actually, never mind. I forgot that the last line will also end in a special character, EOF; as such, `.chomp` will remove that instead.

GA-MEB:
6 of 6

Great job!

PR #42, opened 2016-02-12
Done with respoonse.

dumplingpuff:


lib/diagnostic.rb
```diff
@@ -5,36 +5,40 @@ class Diagnostic
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max
   end
+  # Jason Liang's Comment:
+  # I wanted to write enum.max.length but each entry
+  # implements a length method.
+
```
GA-MEB, line 13: The way to accomplish this is to either 
(a) use `max_by` and pass it a block -- e.g. `enum.max_by { |entry| entry.length }` -- or
(b) use `max` with a custom comparator block built on top of the 'spaceship' operator,
e.g. `enum.max { |a, b| a.length <=> b.length }`

lib/diagnostic.rb
```diff
@@ -5,36 +5,40 @@ class Diagnostic
...
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(:+)
```
GA-MEB, line 20: Correct, but only in Ruby v2.0 and above. In older versions of Ruby, you need to explicitly add the `&` before the symbol:
`enum.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,40 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    enum.any?{ |x| x.odd? }
```
GA-MEB, line 27: Yes. Could also have been written `enum.any?(&:odd?)` / `enum.any?(:odd?)`.

lib/diagnostic.rb
```diff
@@ -5,36 +5,40 @@ class Diagnostic
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    enum.all?{ |x| x.even? }
```
GA-MEB, line 34: Correct; also, see note above.

lib/diagnostic.rb
```diff
@@ -43,6 +47,14 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    lines = File.open(ARGV[0]) do |file|
```
GA-MEB, line 50: We're given the filename as an argument, so there's no need to grab is as `ARGV[0]`.

lib/diagnostic.rb
```diff
@@ -43,6 +47,14 @@ def transform(words)
...
+      file.map(&:itself)
+    end
+    puts lines
   end
 end
+
+# Jason Liang's comments:
+# I copied this directly from the training yesterday
+# This is subject quite unknown to me
+# can you clarify what ARGV[0] means?
```
GA-MEB, line 60: ARGV[0] is a reference to the first argument to the method (similar to `arguments[0]` in JavaScript). No need to use it here, since we have a named first argument.

lib/diagnostic.rb
```diff
@@ -43,6 +47,14 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    lines = File.open(ARGV[0]) do |file|
+      file.map(&:itself)
```
GA-MEB, line 51: `map(&:itself)` just creates a duplicate array, with the same content as the original.
Instead of doing this, we want to get rid of the `\n` at the end of each line and (except for the final one).

GA-MEB:
4.5 / 6 (-0.75 on Q1, Q7)

PR #43, opened 2016-02-12
Ruby Enumerable Diagnostic....

msyao:


lib/diagnostic.rb
```diff
@@ -5,36 +5,37 @@ class Diagnostic
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max(&:size)
```
GA-MEB, line 8: `max` is expecting a block telling it how to compare two entries, using the 'spaceship' operator. 
i.e. `enum.max { |a, b| a.size <=> b.size }`

lib/diagnostic.rb
```diff
@@ -5,36 +5,37 @@ class Diagnostic
...
   end
 
+
   # Question 2
   # Write ruby code to return the sum of entries in the
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(&:+)
```
GA-MEB, line 17: Correct.

lib/diagnostic.rb
```diff
@@ -5,36 +5,37 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.all?(&:odd?)
```
GA-MEB, line 24: Not quite. 

> "if **some** of the entries"

If even a single number is odd, `some_odd` needs to return `true`. You should use `any?` instead of `all?`.

lib/diagnostic.rb
GA-MEB, line 8: This _would_ work if you used `max_by` instead of `max`.

GA-MEB:
3.75 / 6 (-0.75 on Q1, -0.5 on Q3, -1 on Q6)

PR #44, opened 2016-02-12
answer diagnostic questions

kpie89:


lib/diagnostic.rb
```diff
@@ -5,7 +5,7 @@ class Diagnostic
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max_by { |i| i.length }
```
GA-MEB, line 8: Yes, this will work. You can also use `.max` and pass it a custom comparator block:

```
enum.max { |a, b| a.length <=> b.length }
```

lib/diagnostic.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(:+)
```
`enum.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |i| i.odd? }
```
GA-MEB, line 23: Yep. You can also use syntax like in Q2: `numbers.any?(:&odd)` / `numbers.any?(:odd?)`

lib/diagnostic.rb
```diff
@@ -43,6 +43,10 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    File.open('filename') do |file|
+  file.each_line do |line|
+    puts line
+   end
```
GA-MEB, line 49: You almost opened the file correctly, but you didn't do the correct behavior with each line.

> "Concatenate the lines together after replacing the trailing newline character with a        space and return the result."

lib/diagnostic.rb
```diff
@@ -43,6 +43,10 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    File.open('filename') do |file|
```
GA-MEB, line 46: Not `'filename'` (string), `filename` (variable).

GA-MEB:
5.25 / 6 (-0.75 on Q7)

PR #45, opened 2016-02-12
initial diagnostic commit

jenweber:
Ran out of time

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -5,7 +5,7 @@ class Diagnostic
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max { |a, b| a.length <=> b.length }
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -5,7 +5,7 @@ class Diagnostic
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max { |a, b| a.length <=> b.length }
```
GA-MEB, line 8: Correct! You could also use `max_by`, like so:

lib/diagnostic.rb
```diff
@@ -13,28 +13,32 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(:+)
```
`enum.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -13,28 +13,32 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    x = false
+    numbers.each {|num| num.odd? ? x = true : x }
+    return x
```
GA-MEB, line 25: This will work. However, the Enumerable module gives us a more elegant way to handle this: `numbers.any? { |num| num.odd? }`
This can also be shortened as `numbers.any?(&:odd?)` / `numbers.any?(:odd)`

lib/diagnostic.rb
```diff
@@ -13,28 +13,32 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    x=true
+    numbers.each {|num| num.even? ? x : x = false }
+    return x
```
GA-MEB, line 34: As above, this will work; however (since this is, after all, a diagnostic about Enumerables) a better approach might have been `numbers.all? { |num| num.even? }`

This can also be written `numbers.all?(&:even?)` / `numbers.all?(:even?)`

lib/diagnostic.rb
```diff
@@ -13,28 +13,32 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.each {|word| word.capitalized}
```
GA-MEB, line 41: The method is `capitalize`, not `capitalized`.

GA-MEB, line 41: This also will not return an array, because the `.each` method returns `undefined`. The right tool to use here is `.map`.

GA-MEB:
3.75 / 6 (-0.25 on Q3 and Q4, -0.75 on Q5, -1 on Q6)

PR #46, opened 2016-02-12
diagnostic response

MFBrewster:


lib/diagnostic.rb
```diff
@@ -13,28 +17,32 @@ def longest(enum)
...
+    end
+    sum
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any?  { |x| x.odd? }
```
GA-MEB, line 31: Correct. Could also have been written `numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -13,28 +17,32 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |x| x.even? }
```
GA-MEB, line 38: Correct. Could also have been written `numbers.all?(&:even?)`

lib/diagnostic.rb
```diff
@@ -13,28 +17,32 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map { |word| word.upcase }
```
GA-MEB, line 45: `upcase` make it all caps - the right method here is `capitalize`. Otherwise correct.

GA-MEB, line 45: Could also have been written `words.map(&:capitalize)`

lib/diagnostic.rb
```diff
@@ -13,28 +17,32 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    sum = 0;
+    enum.foreach() do |num|
```
GA-MEB, line 21: `forEach` is JavaScript. If you'd used `.each`, this code would work.

GA-MEB, line 21: The better answer for this question is to use `.reduce`:
`enum.reduce { |memo, num| memo + num }` (which could also be written `enum.reduce(&:+)`)

lib/diagnostic.rb
```diff
@@ -5,7 +5,11 @@ class Diagnostic
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    long_entry = ''
+    enum.foreach() do |one_line|
```
GA-MEB, line 9: `forEach` is JavaScript. If you'd used `.each`, this code would work.

GA-MEB, line 9: The better answer here would have been to use a comparison method from Enumerable, like `max` or `max_by`.
e.g.
`enum.max_by(&:length)`

lib/diagnostic.rb
```diff
@@ -43,6 +51,9 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    filename.each_with_object(new_string: '') do |line|
```
GA-MEB, line 54: `filename` is just like it sounds - the name of the file. It is not the file itself.

lib/diagnostic.rb
```diff
@@ -43,6 +51,9 @@ def transform(words)
...
+      new_string += line.gsub(/\n/, ' ')
```
GA-MEB, line 55: This `gsub` step would be effective for operating on an individual line.

lib/diagnostic.rb
```diff
@@ -43,6 +51,9 @@ def transform(words)
...
+    end
+    new_string
```
GA-MEB, line 57: Here's one potential solution:
`File.open(filename).map { | line | line.gsub(/\n/, ' ') }.reduce(&:+)`

GA-MEB:
3.75 / 6 (-0.5 on Q1 and Q2; -0.25 on Q5; -1 on Q7)

PR #47, opened 2016-02-12
Ajordon diag response

ajordon:


lib/diagnostic.rb
```diff
@@ -6,6 +6,10 @@ class Diagnostic
   # method that returns a Fixnum.
   def longest(enum)
     # Your code goes here
+    longest = enum.reduce do |memo, obj|
+      memo.length > obj.length ? memo : obj
+    end
```
GA-MEB, line 11: This is one of several ways that you could get the desired behavior. However, there are easier options: for instance, `max` or `max_by`. Here's how they get called.
`enum.max_by { |entry| entry.length }` or `enum.max_by(&:length)`
`enum.max { |a.b| a.length <=> b.length }`

lib/diagnostic.rb
```diff
@@ -6,6 +6,10 @@ class Diagnostic
...
+    p longest
```
GA-MEB, line 12: Return, not print. In fact, `print` returns `undefined`, so this method will return `undefined`.

lib/diagnostic.rb
```diff
@@ -14,6 +18,8 @@ def longest(enum)
   # Assume each entry implements a `+` method
   def sum(enum)
     # Your code goes here
+    total = enum.reduce(:+)
```
GA-MEB, line 21: Correct.

lib/diagnostic.rb
```diff
@@ -14,6 +18,8 @@ def longest(enum)
...
+    p total
```
GA-MEB, line 22: See above.

lib/diagnostic.rb
```diff
@@ -21,20 +27,26 @@ def sum(enum)
   # in the numbers enumerable are odd
   def some_odd(numbers)
     # Your code goes here
+   numbers.one? { |num| num.odd? }
```
GA-MEB, line 30: This will return `true` if _exactly one_ of the entries is odd. What if there were two?
The right tool to use in this scenario is `.any?`

lib/diagnostic.rb
```diff
@@ -21,20 +27,26 @@ def sum(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    evens = numbers.all { |num| num.even? }
```
GA-MEB, line 37: Should be `.all?`, but otherwise right.

lib/diagnostic.rb
```diff
@@ -21,20 +27,26 @@ def sum(enum)
...
+    puts evens
```
GA-MEB, line 38: See previous notes.

lib/diagnostic.rb
```diff
@@ -21,20 +27,26 @@ def sum(enum)
...
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    lines = filename.open(ARGV[0]) do |file|
+      file.map(&:itself).capitalize
+    end
+
+    print lines
```
GA-MEB, line 49: No, we're not opening a file in this example. We were just looking for a simple `.map` here.
e.g. `words.map(&:capitalize)`

lib/diagnostic.rb
```diff
@@ -43,6 +55,9 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
-  end
+    lines = filename.open(ARGV[0]) do |file|
+      file.map(&:itself)
+    end
+
+  print lines
```
GA-MEB, line 62: There's no need to reference `ARGV[0]` here; since our first parameter is named, we can refer to it by name.

GA-MEB, line 62: `map(&:itself)` just creates a duplicate array, with the same content as the original.
Instead of doing this, we want to get rid of the `\n` at the end of each line and (except for the final one).

GA-MEB:
3 / 6 (-0.25 on Q1 and Q2; -0.5 on Q3 and Q4; -0.75 on Q5; -0.75 on Q6)

PR #48, opened 2016-02-12
answer some

byau:


lib/diagnostic.rb
```diff
@@ -1,10 +1,12 @@
 # Ruby enumerable diagnostic
 class Diagnostic
+
   # Question 1
   # Write ruby code to return the longest entry in the
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
+    enum.reduce(0) { |m, o| m = m.length < o.length? o : m }
```
GA-MEB, line 9: This will _almost_ work - passing in 0 as a starting value will actually break the `reduce`, since `0` doesn't have a `.length` method. If you didn't include a starting value, it would just begin with the entry 0 and entry 1.

GA-MEB, line 9: Enumerable provides some simpler methods for doing this kind of work: `max` and `max_by`.
e.g.

lib/diagnostic.rb
```diff
@@ -13,6 +15,7 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
+    enum.reduce(0) { |m, o| m += m}
```
GA-MEB, line 18: Correct. This could also be written more simply as `enum.reduce(&:+)` (or just `enum.reduce(:+)` in Ruby 2.0 and above).

lib/diagnostic.rb
```diff
@@ -20,6 +23,8 @@ def sum(enum)
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
+    enum.each_with_object(false) { |o, m| m = true if o.odd? }
+    m
```
GA-MEB, line 27: Almost! `m` doesn't exist outside of the block above, so having that be the last line causes the method to return `undefined`. If you had just left that line out, it would have worked, since `.each_with_object` returns the modified object as a result.

GA-MEB, line 27: A simpler way to do this would be to use `.any?`

lib/diagnostic.rb
```diff
@@ -27,6 +32,8 @@ def some_odd(numbers)
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
+    enum.each_with_object(true) { |o, m| m = false if o.odd? }
+    m
```
GA-MEB, line 36: See comments from previous question. The simpler way to do this would be to use `.all?`

lib/diagnostic.rb
```diff
@@ -34,6 +41,8 @@ def every_even(numbers)
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
+    enum.each_with_object([]) { |o, m| m << o.capitalized }
+    m
```
GA-MEB, line 45: See comments above. A simpler way to do this would have been to use `.map`.

GA-MEB:
4 / 6 (-0.25 on Q1, Q3, Q4, and Q5; -1 on Q6)

PR #49, opened 2016-02-12
Celena Diagnostic Response

seatuna:


lib/diagnostic.rb
```diff
@@ -5,7 +5,7 @@ class Diagnostic
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.reduce { |memo, obj| memo.length > obj.length ? memo : obj }
```
GA-MEB, line 8: Correct. For your reference, though, Enumerable provides some simpler methods for doing this kind of work: `max` and `max_by`.
e.g.

lib/diagnostic.rb
```diff
@@ -13,28 +13,29 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(:+)
```
`enum.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -13,28 +13,29 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.odd?
```
GA-MEB, line 23: Not quite. Fixnums have a `.odd?` method; arrays don't. Here's one solution:
`numbers.any? { |num| num.odd? }` 
(alternatively, `numbers.any?(&:odd?)`)

lib/diagnostic.rb
```diff
@@ -13,28 +13,29 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    array = enum.each_with_object([]) { numbers.even? ? true : false }
+    array.include?(false) ? false : true
```
GA-MEB, line 31: `array` should be `numbers`. No reference inside the block to the given number.

Also, that `array.include?` line could be simplified as
`!array.include?(false)`

GA-MEB, line 31: A better approach would have been to use `.all?`:
`numbers.all? { |num| num.even? }`
(alternatively, `numbers.all?(&:even?)`)

lib/diagnostic.rb
```diff
@@ -13,28 +13,29 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.each_with_object([]) { words.capitalize! }
```
GA-MEB, line 38: No reference inside the block to the given word.
Also, why use `each_with_object` if you're just going to operate in place?
`words.each { |word| word.capitalize! }`

GA-MEB, line 38: A better tool for this job would be `.map` :
`words.map { |word| word.capitalize }`
(alternatively, `words.map(&:capitalize)`)

lib/diagnostic.rb
```diff
@@ -43,6 +44,10 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    lines = File.open(ARGV[0]) do |file|
```
GA-MEB, line 47: ARGV[0] is a reference to the first argument to the method (similar to `arguments[0]` in JavaScript). No need to use it here, since we have a named first argument.

lib/diagnostic.rb
```diff
@@ -43,6 +44,10 @@ def transform(words)
...
+      file.map(&:itself)
```
GA-MEB, line 48: This creates a duplicate collection of lines, but it doesn't perform the specified transformation:

> "Concatenate the lines together after replacing the trailing newline character with a space and return the result."

lib/diagnostic.rb
```diff
@@ -43,6 +44,10 @@ def transform(words)
...
+    end
+
+    puts lines
```
GA-MEB, line 51: Should return (not print) the result.

GA-MEB:
3.5 / 6 (-0.75 on Q3; -0.5 on Q4 and Q5; -0.75 on Q6)

PR #50, opened 2016-02-12
James Friedman Response

jpfriedman18:
Couldn't get last test to pass

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -5,7 +5,7 @@ class Diagnostic
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max { |a, b| a.length <=> b.length }
   end
 
   # Question 2
@@ -13,28 +13,28 @@ class Diagnostic
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(:+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |num| num.odd? }
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |num| num.even? }
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map { |word| word.capitalize }
   end
 
   # Question 6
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -5,7 +5,7 @@ class Diagnostic
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max { |a, b| a.length <=> b.length }
```
GA-MEB, line 8: Correct. `.max_by(&:length)` would also work.

lib/diagnostic.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(:+)
```
`enum.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |num| num.odd? }
```
GA-MEB, line 23: Correct. Equivalently written `numbers.any?(&:odd?)` / `numbers.any?(:odd?)`

lib/diagnostic.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |num| num.even? }
```
GA-MEB, line 30: Correct. Equivalently written `numbers.all?(&:even?)` / `numbers.all?(:even?)`

lib/diagnostic.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map { |word| word.capitalize }
```
GA-MEB, line 37: Correct. Equivalently written `words.map(&:capitalize)` / `words.map(:capitalize)`

lib/diagnostic.rb
```diff
@@ -43,6 +43,11 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    lines = File.open(filename) do |file|
+      file.map(&:itself)
+    end
+    normalized_lines = lines.each {|line| line.chomp('\n') }
+    concatenated_lines = normalized_lines.join(' ')
+    return concatenated_lines
```
GA-MEB, line 51: VERY close, and to be honest, I'll take it. The official solution says to strip trailing whitespace, but that's not clearly indicated in the question.

GA-MEB, line 51: Some of these steps could be simplified and combined like so:
`File.open(filename).map { | line |  line.chomp + ' ' }.reduce(&:+).strip`
The official solution is even shorter:
`File.open(filename).reduce('') { |a, e| a << e.chomp + ' ' }.strip`

GA-MEB:
6 / 6
Nice work.

PR #51, opened 2016-02-12
response

rchezrutt88:


lib/diagnostic.rb
```diff
@@ -5,7 +5,7 @@ class Diagnostic
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max_by {|x| x.length}
```
GA-MEB, line 8: Correct. This could also have been written `enum.max_by(&:length)`.

GA-MEB, line 8: Yet another solution would be to use `.max` :

lib/diagnostic.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(:+)
```
`enum.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? {|x| x.odd?}
```
GA-MEB, line 23: Correct. Equivalently written `numbers.any?(&:odd?)` / `numbers.any?(:odd?)`

lib/diagnostic.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? {|x| x.even?}
```
GA-MEB, line 30: Correct. Equivalently written `numbers.all?(&:even?)` / `numbers.all?(:even?)`

lib/diagnostic.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map {|word| word.capitalize}
```
GA-MEB, line 37: Correct. Equivalently written `words.map(&:capitalize)` / `words.map(:capitalize)`

lib/diagnostic.rb
```diff
@@ -43,6 +43,7 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    # file = File.new(filename)
+    # file.reduce {|memo, line| memo = line.sub(\n, )}
```
GA-MEB, line 47: Here's the official solution for this one.
`File.open(filename).reduce('') { |a, e| a << e.chomp + ' ' }.strip`

GA-MEB:
5 / 6 (-1 on Q7)

PR #52, opened 2016-02-12
ruby enumerable diagnostic

apalmer0:


lib/diagnostic.rb
```diff
@@ -5,7 +5,13 @@ class Diagnostic
...
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    longest = enum.first
+    enum.each do |entry|
+      if entry.length > longest.length
+        longest = entry
+      end
+    end
+    longest
```
GA-MEB, line 14: This will work, but Enumerable specifically provides us with tools to address situations like this: `max` and `max_by`.
e.g.
`enum.max_by { |entry| entry.length }` or `enum.max_by(&:length)`

lib/diagnostic.rb
```diff
@@ -13,28 +19,35 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(0) { |m, p| m += p }
```
GA-MEB, line 22: Correct. This could also have been written `enum.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -13,28 +19,35 @@ def longest(enum)
...
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.each do |num|
+      if num.odd?
+        return true
+      end
+    end
```
GA-MEB, line 33: This will do the desired behavior. But as in Q1, there's an easier way : `.any?`
e.g.
`numbers.any? { |num| num.odd? }` or `numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -13,28 +19,35 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    sum = numbers.reduce(0) { |memo, obj| memo += obj }
+    sum % 2 == 0
```
GA-MEB, line 41: This works, but the solution is a bit 'clever' - it's not super applicable in different circumstances (e.g. if we'd asked for them all to be odd, or square).

Enumberable's `.all?` method, in contrast, is very versatile.
Consider the following cases.
`numbers.all?(&:even?)` 
`numbers.all?(&:odd?)`

```
def is_square? (num)
    ... 
end
numbers.all? { |num| is_square?(num) }
```

lib/diagnostic.rb
```diff
@@ -13,28 +19,35 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.each do |word|
+      word.capitalize!
+    end
```
GA-MEB, line 50: This will work. You could write it more simply as `words.each(&:capitalize!)`.

A better approach, though, would be to use `.map`, because operating in place (`!`) is a dangerous practice.
`words.map { |word| word.capitalize }` ( or `words.map(&:capitalize)`)

lib/diagnostic.rb
```diff
@@ -43,6 +56,8 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    File.open(filename) do |file|
+      file.split('\n').join()
+    end
```
GA-MEB, line 61: You're thinking along the right lines, but that's not how `File.open` works. When you give a block to `File.open`, the thing that the block will be passed is a **line of text** from the file, not the entire contents.

`File.open(filename).map { | line | line.chomp }.join(' ').reduce(&:+).strip` is one way to solve this.

GA-MEB:
5.5 / 6 (-0.5 on Q7)

I didn't take off for any of the alternative implementations, but you should really try to get acquainted with some of the methods that Enumerable provides - they're more semantic, and simpler to write.

PR #53, opened 2016-02-12
Dia complete

edesilets:


Gemfile.lock
```diff
@@ -1,51 +1,50 @@
...
-  pry (~> 0.10.1)
+  byebug (~> 6.0)
+  pry (~> 0.10)
   rspec (~> 3.3)
-  rubocop (~> 0.34.2)
+  rubocop (~> 0.36)
 
 BUNDLED WITH
-   1.10.6
+   1.11.2
```
GA-MEB, line 50: What happened here??

Did you fork an earlier version of this repo, before it was shared out?

lib/diagnostic.rb
```diff
@@ -1,11 +1,11 @@
...
-class Assessment
+# Ruby enumerable diagnostic
+class Diagnostic
   # Question 1
   # Write ruby code to return the longest entry in the
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.length!
```
GA-MEB, line 8: Sorry, this isn't right. We want to find the element with the longest length, not the size of the collection.

lib/diagnostic.rb
```diff
@@ -13,28 +13,31 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.each do |prev_item, nex_item|
```
GA-MEB, line 16: `.each` doesn't give blocks two arguments, only one - the element that we're currently looking at.

lib/diagnostic.rb
```diff
@@ -13,28 +13,31 @@ def longest(enum)
...
+      prev_item += nex_item
+    end
```
GA-MEB, line 18: You're spinning your wheels with this. `prev_item` is not defined outside of the block, so data isn't being stored from one iteration to the next, and `.each` returns `undefined`, so the method will return `undefined` in any case.

lib/diagnostic.rb
```diff
@@ -13,28 +13,31 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.odd?
```
GA-MEB, line 25: `.odd?` is a method on a number, not a collection.

lib/diagnostic.rb
```diff
@@ -13,28 +13,31 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.even?
```
GA-MEB, line 32: See previous.

lib/diagnostic.rb
```diff
@@ -13,28 +13,31 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
-  # entries from the words enumerable capitalized.
+  # entries from the words enumerable capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
     # Your code goes here
+    words.upcase!
```
GA-MEB, line 40: `.upcase!` is a method that Strings have, not collections.

lib/diagnostic.rb
```diff
@@ -43,6 +46,10 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    File.open(filename) do |file|
+    file.each do |line|
+      
+      #replace trailing new line char "\n" with " "
+    end
```
GA-MEB, line 53: `File.open(filename)` is correct. However, the input parameter to the block is actually a single line of text from the file, rather than the entire contents.

[EDIT] Whoops - please disregard this comment. I missed that you were giving the `do` block directly to the file, rather than trying to use `.each`. This setup is actually correct. However, the `do` block is not necessary - we could just call `.each` on `File.open(filename)`.

GA-MEB:
We should chat about this one. Would you have time for a 1:1 this week?

PR #54, opened 2016-02-12
wrote some answers

Knguyen21:
the last two code messes things up

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -5,7 +5,7 @@ class Diagnostic
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max { |a, b| a.length <=> b.length }
   end
 
   # Question 2
@@ -13,28 +13,28 @@ class Diagnostic
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    (enum).inject { |sum, n| sum + n }
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
-  def some_odd(numbers)
-    # Your code goes here
+  def some_odd(numbers)   # one or more
+    numbers.any? { |num| num % 2 != 0 }
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |num| num % 2 == 0}
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    (words).each_with_object(%w[]) { |i, a| a.capitalized }
   end
 
   # Question 6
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -5,7 +5,7 @@ class Diagnostic
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max { |a, b| a.length <=> b.length }
```
GA-MEB, line 8: Yep. `enum.max_by { |entry| entry.length}` is another valid approach.

lib/diagnostic.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    (enum).inject { |sum, n| sum + n }
```
GA-MEB, line 16: Correct. 
BTW, `inject` and `reduce` are actually the exact same method in Ruby - one is an alias for the other.

lib/diagnostic.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
-  def some_odd(numbers)
-    # Your code goes here
+  def some_odd(numbers)   # one or more
+    numbers.any? { |num| num % 2 != 0 }
```
GA-MEB, line 23: Correct. Numbers actually have a method that gives whether they are odd, `.odd?`, so you could rewrite this as `numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |num| num % 2 == 0}
```
GA-MEB, line 30: Correct, but as above, there's a method you could have used as a shortcut, `.even?`
e.g.
`numbers.all?(&:even?)`

lib/diagnostic.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    (words).each_with_object(%w[]) { |i, a| a.capitalized }
```
GA-MEB, line 37: The method is `capitalize`, not `capitalized`.

GA-MEB, line 37: `%w[]` is not proper syntax. You could have just written `[]`.

GA-MEB, line 37: This solution will work, but a better solution would be to use `.map`
e.g.
`words.map(&:capitalize)`

lib/diagnostic.rb
```diff
@@ -43,6 +43,10 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    File.open(filename) do |file|
+      file.each do |line, memo|
+          line.gsub!(/\n/, " ")
+          p line
+        end
+    end
```
GA-MEB, line 51: `File.open(filename)` gives us the single file that we're loading. The File class is Enumerable, so we can call Enumerable methods on files.

GA-MEB, line 51: The `.each` block won't do what you think. Nothing gets returned by `p`.

lib/diagnostic.rb
GA-MEB, line 37: What is the significance of `i` and `a` as parameter names?

GA-MEB:
5.25 / 6 (-0.25 on Q5, -0.5 on Q6)
Nice work.

PR #55, opened 2016-02-12
nw resp

nwilcox11:


lib/diagnostic.rb
```diff
@@ -5,7 +5,7 @@ class Diagnostic
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+
```
GA-MEB, line 8: ?

lib/diagnostic.rb
```diff
@@ -13,28 +13,29 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce do |sum, n|
+      sum + n
+    end
```
GA-MEB, line 18: Correct. You could also have written this as `enum.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -13,28 +13,29 @@ def longest(enum)
...
   def some_odd(numbers)
-    # Your code goes here
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.even?
```
GA-MEB, line 31: `.even?` is a method on a number - it's not included in the Enumerable module.

GA-MEB, line 31: What you want to be doing here is checking each number and running `.even?` on it. 
In particular, since you want to check if every number is even, you could write:
`numbers.all? { |num| num.even? }` (could also be written as `numbers.all?(&:even?)`)

lib/diagnostic.rb
```diff
@@ -43,6 +44,10 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+      lines = File.open(filename[0]) do |file|
+      file.each_with_object(lines: 0, new_lines: 0) do |line, memo|
+        memo[:lines] += line.split()
+        memo[:new_lines] += line.join(' ')
+      end
```
GA-MEB, line 51: This answer is kinda out in left field.

GA-MEB, line 51: To load a file, you run `File.open(filename)`. This gives you a new File object. The File class includes the Enumerable module, so Files have all of the Enumerable methods, such as `.map`. `.map` accepts a block, and passes that block a single item (in this case, a line of text from the file) as it iterates through the collection.

GA-MEB, line 51: Here's one possible solution:
`File.open(filename).map{ | line | line.chomp}.join(' ').reduce(&:+).strip`
The `.strip` at the end is to get rid of any trailing whitespace we may have accidentally included.

GA-MEB:
We should chat about this one. Would you have time for a 1:1 this week?

PR #56, opened 2016-02-12
response to diagnostic

jdamiron:


lib/assessment.rb
```diff
@@ -5,7 +5,7 @@ class Assessment
   # enumerable supplied.  Assume each entry implements a length
   # method that returns a Fixnum.
   def longest(enum)
-    # Your code goes here
+    enum.max_by { |x| x.length  }
```
GA-MEB, line 8: Correct. This could also have been written `enum.max_by(&:length)`

GA-MEB, line 8: Another possible solution would be to use `.max`:
`enum.max { |a, b| a.length <=> b.length}`

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
   # enumerable supplied (i.e. apply the `+` operator)
   # Assume each entry implements a `+` method
   def sum(enum)
-    # Your code goes here
+    enum.reduce(:+)
```
GA-MEB, line 16: Correct, but only in Ruby v2.0 and above. In older versions of Ruby, you need to explicitly add the `&` before the symbol:
`enum.reduce(&:+)`

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the numbers enumerable are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |n| n.odd? }
```
GA-MEB, line 23: Correct. Could also have been written `numbers.any?(&:odd?)`

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # numbers enumerable are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |n| n.even? }
```
GA-MEB, line 30: Correct. Could also have been written `numbers.all?(&:even?)`

lib/assessment.rb
```diff
@@ -13,28 +13,28 @@ def longest(enum)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the words enumerable capitalized.
   def transform(words)
-    # Your code goes here
+    words.map { |words| words.capitalize }
```
GA-MEB, line 37: Correct. Could also have been written `words.map(&:capitalize)`

lib/assessment.rb
```diff
@@ -43,6 +43,8 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    File.read(filename) do |line|
+
+    end
```
GA-MEB, line 48: You've loaded the file correctly, but your setup with the block is slightly off. Files are Enumerable, so they can call `.each`, `.map`, etc. If you called one of those methods, the block you'd pass it would accept a single line as the argument. However, if you just give `File.open(filename)` a `do` block, it will be handed the file itself.

GA-MEB:
5.25 / 6 (-0.75 on Q6)
Nice work.

PR #68, opened 2016-04-15
Answered Questions in Diagnostic

louarnos:


lib/diagnostic.rb
```diff
@@ -43,6 +43,12 @@ def transform(words)
...
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    output = ''
+    filename.open(ARGV[0]) do |file|
+      file.each do |line|
+        line.to_a.pop.join
+        output = output + line
+      end
+    output
```
RDegnen, line 52:
``` rb
def read_file(filename)
  end
```

RDegnen:
5 of 6 nice!

PR #70, opened 2016-04-15
Complete ruby enumerable diagnostic

MichaelB167:


lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.reduce { |a, e| a.length > e.length ? a : b}
```
RDegnen, line 8: `list.max { |a, b| a.length <=> b.length }`

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
-    # Your code goes here
+    list.reduce(:+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |num| num.odd}
```
RDegnen, line 23: so close, you just needed to add ? onto odd. so `odd?`

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |num| num.even}
```
RDegnen, line 30: same as above

lib/diagnostic.rb
```diff
@@ -43,6 +43,11 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    filename.open(ARGV[0]) do |file|
+      file.each_with_index do |line, i,|
+        puts "#{i}: #{line}"
+        #I assume "\n" is meant to go somewhere but not sure where
+      end
+    end
```
RDegnen, line 51: `def read_file(filename)
  end
`

RDegnen:
3 of 6. I gave you half points on the two you came super close on

PR #72, opened 2016-06-13
Response

MishaHerscu:
Comfort: 4
Clarity: 4

So far this has been from memory. I will likely use google/ruby documentation and my class notes and the class repos for the other questions.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -15,7 +15,7 @@ class Diagnostic
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
   end
 
   # Question 3
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -29,7 +29,7 @@ class Diagnostic
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all?(&:even?)
   end
 
   # Question 5
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -36,7 +36,7 @@ class Diagnostic
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map { |word| word[0].upcase + word.slice(1, word.length - 1) }
   end
 
   # Question 6
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -51,6 +51,6 @@ class Diagnostic
         contents += line.tr("\n", ' ')
       end
     end
-    contents
+    contents.chop
   end
 end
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -5,36 +5,38 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    lengths = []
+    list.each { |element| lengths.push(element.length) }
+    list[lengths.index(lengths.max)]
```
berziiii, line 10: This may work, but since the question wanted a comparison, we should use the spaceship operator

lib/diagnostic.rb
```diff
@@ -5,36 +5,38 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
```
berziiii, line 18: Correct.

lib/diagnostic.rb
```diff
@@ -5,36 +5,38 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any?(&:odd?)
```
berziiii, line 25: Correct.

lib/diagnostic.rb
```diff
@@ -5,36 +5,38 @@ class Diagnostic
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all?(&:even?)
```
berziiii, line 32: Correct.

lib/diagnostic.rb
```diff
@@ -5,36 +5,38 @@ class Diagnostic
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map { |word| word[0].upcase + word.slice(1, word.length - 1) }
```
berziiii, line 39: This may work, but even easier

`words.map(&:capitalize)`

lib/diagnostic.rb
```diff
@@ -43,6 +45,12 @@ def transform(words)
...
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    contents = ''
+    File.open(filename) do |file|
+      file.each do |line|
+        contents += line.tr("\n", ' ')
+      end
+    end
+    contents.chop
```
berziiii, line 54: May work. I'll give you it.

berziiii:
6 / 6. Remember that enumberables are made to make our lives easier.

lib/diagnostic.rb
MishaHerscu, line 54: what is the better way to do this last one?

PR #73, opened 2016-06-13
ruby diagnostic

jacobhereford:
answered all questions accordingly

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -5,37 +5,73 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    ar = ['one','two','three','four','five']
+      ar.max_by(&:length) # => "three"
   end
+
+  Clarity:2
+  Comfort:2
+
+  Googled answer
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    [1, 2, 3].inject(0) {|sum, i|  sum + i }
+
   end
+
+  Clarity:3
+  Comfort:3
+
+  Googled answer
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    100.times do |i|
+    next if i % 2 != 0
+    puts i
+    end
   end
+
+  Clarity:3
+  Comfort:3
+
+  Googled answer
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    100.times do |i|
+    next if i % 2 == 0
+    puts i
+    end
   end
+
+  Clarity:3
+  Comfort:3
+
+  Googled answer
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -5,44 +5,87 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    ar = ['one','two','three','four','five']
+      ar.max_by(&:length) # => "three"
```
berziiii, line 9: You need to have a comparison so remember to use the spaceship operator.

lib/diagnostic.rb
```diff
@@ -5,44 +5,87 @@ class Diagnostic
...
+
+  Googled answer
+
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    [1, 2, 3].inject(0) {|sum, i|  sum + i }
```
berziiii, line 22: Even simpler.

`list.reduce(:+)`

lib/diagnostic.rb
```diff
@@ -5,44 +5,87 @@ class Diagnostic
...
+
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    100.times do |i|
+    next if i % 2 != 0
+    puts i
+    end
```
berziiii, line 38: Remember that enumerable methods are used to make our lives easier

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -5,44 +5,87 @@ class Diagnostic
...
+  Comfort:3
+
+  Googled answer
+
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    100.times do |i|
```
berziiii, line 50: Same here.

`numbers.any?(&:even?)`

The question also wants a true or false returned, not the number itself.

lib/diagnostic.rb
```diff
@@ -5,44 +5,87 @@ class Diagnostic
...
+  Comfort:3
+
+  Googled answer
+
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    manufacturer.MFA_BRAND.first.upcase
```
berziiii, line 65: HA, not sure where this came from.

`words.map(&:capitaize)`

lib/diagnostic.rb
```diff
@@ -5,44 +5,87 @@ class Diagnostic
...
+
   # Question 6
   # Write ruby code to read each line from the file
   # named filename. Concatenate the lines together after
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    [filename].concat do
+      puts.line + " "
```
berziiii, line 83:
``` ruby
  file.each do |line|
  line.chomp.join(' ')
  end
end
```

Something along this path

berziiii:
I'll give you 2 / 6 but remember to read the questions! If they ask for a true or false, then you need to use the methods that return a boolean and not the result itself

PR #74, opened 2016-06-13
response

huey1021:
comfort-3
clarity- 3

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -5,7 +5,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    longest.max_by(&:length)
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    longest.max_by(&:length)
```
berziiii, line 8: Since it's a comparison, we should use the spaceship operator.

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    sum.reduce(:+)
```

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+  some_odd.select(&:odd)
```
berziiii, line 23: You want a true or false return so you have to use `.any`

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    return every_even.all?(&:even?)
```
berziiii, line 30: Same here

`numbers.any?(&:even?)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.split.map(&:capitalize)
```
berziiii, line 37: Almost. no split needed. You can just call

`words.map(&:capitalize)`

lib/diagnostic.rb
```diff
@@ -43,6 +43,6 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    puts File.read(filename)\n
```
berziiii, line 46:
``` ruby
  file.each do |line|
   line.chomp.join(' ')
  end
end
```

something along these lines should work.

berziiii:
3 / 6. Nice work. Small mistakes but right approach.

PR #75, opened 2016-06-13
questions answered except last one

tfavazza:
Comfort: 3
Clarity: 4

I used Google, Ruby docs, and rake test to help finish this assignment

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -5,7 +5,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max_by(&:length)
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max_by(&:length)
```
berziiii, line 8: Since its a comparison, we should use the spaceship operator. 

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
```

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any?(&:odd?)
```
berziiii, line 23: Correct.

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all?(&:even?)
```
berziiii, line 30: Correct.

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map(&:capitalize)
```
berziiii, line 37: Correct.

lib/diagnostic.rb
```diff
@@ -43,6 +43,6 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    File.readlines(filename).chomp.join(' ')
```
berziiii, line 46: This may work, but the instructions as for the File.open tactic

``` ruby
  file.each do |line|
   line.chomp.join(' ' )
  end
end
```

something along these lines

berziiii:
5 / 6. Nice work!

PR #76, opened 2016-06-13
Ruby enumerable diagnostic

RoobScoobs:
complete questions 1 - 6

Comfort: 3
Clarity: 3

I used the class notes that resembled similar examples and the Ruby documentation to find the enumerable methods.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -5,7 +5,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max_by(&:length)
   end
 
   # Question 2
@@ -13,28 +13,28 @@ class Diagnostic
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.select(&:odd?)
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.select(&:even?)
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map(&:upcase)
   end
 
   # Question 6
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max_by(&:length)
```
berziiii, line 8: May work. But you need to use the comparison operator since it's a comparison. 

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
```

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.select(&:odd?)
```
berziiii, line 23: Almost. 

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.select(&:even?)
```
berziiii, line 30: select would get the value, you want to return true or false

`numbers.any?(&:even?)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map(&:upcase)
```
berziiii, line 37: Yes. but I don't know if `upcase` works.

`words.map(&:capitalize)`

lib/diagnostic.rb
```diff
@@ -43,6 +43,9 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    File.open('filename') do |file|
+      file.each do |line|
+    puts line
```
berziiii, line 48: You're so close! You have to use `.chomp` and + ' ' to alter the lines in the file.

berziiii:
2 / 6. Very close on several. Keep up the good hard work and let us know if you need more review of enumberables.

PR #77, opened 2016-06-13
Ruby enumerable diagnostic

lcurran:
Q6 incomplete

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -5,7 +5,12 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    max = 0
+    list.each { |item|
+      if item.length > max.length
+        max = item
+      }
+      puts max
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -5,36 +5,41 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    max = 0
+    list.each { |item|
+      if item.length > max.length
```
berziiii, line 10: Yes this would work. 

But remember the spaceship comparison

`list.max { |a , b| a.length <=> b.length }`

lib/diagnostic.rb
```diff
@@ -5,36 +5,41 @@ class Diagnostic
...
+      puts max
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
```
berziiii, line 21: Correct.

lib/diagnostic.rb
```diff
@@ -5,36 +5,41 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any?{ |num| num.odd? }
```
berziiii, line 28: Correct. but even shorter hand

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,41 @@ class Diagnostic
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.none?{ |num| num.odd? }
```
berziiii, line 35: Looks like ti would work, but remember short hand

`numbers.any?(&:even?)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,41 @@ class Diagnostic
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.each{ |word| word.capitalize }
```
berziiii, line 42: You would need to change the array. 

`words.map(&:capitalize)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,41 @@ class Diagnostic
...
   end
 
   # Question 6
```
berziiii, line 45:
``` ruby
  file.each do |line|
   line.chomp + ' '
  end
end
```

Something along these lines

berziiii:
4 / 6. Nice work.

PR #79, opened 2016-06-13
Jon J ruby ennumerable diagnostic

modularjon:
Comfort: 4
Clarity: 4

Used ruby documentation and previous ruby io enumerables lesson to answer
questions

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -5,7 +5,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max
   end
 
   # Question 2
@@ -13,28 +13,28 @@ class Diagnostic
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |number| number % 2 != 0 }
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |number| number % 2 == 0 }
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map { |obj| obj.capitalize }
   end
 
   # Question 6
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max
```
berziiii, line 8: Almost. You need to use the spaceship comparison to see what is the largest.

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
```

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |number| number % 2 != 0 }
```
berziiii, line 23: Correct. but think short hand. Using an enumerable method.

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |number| number % 2 == 0 }
```
berziiii, line 30: Same here But yes this would work

`numbers.any?(&:even?)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,36 @@ class Diagnostic
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map { |obj| obj.capitalize }
```
berziiii, line 37: Again here. Correct but think smaller
`words.map(&:capitalize)`

lib/diagnostic.rb
```diff
@@ -43,6 +43,6 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    filename.readlines(filename).map { |obj| obj << ' ' } .reduce(:+)
```
berziiii, line 46: Good thought but you have to use `.chomp`

``` ruby
  file.each do |line|
    line.chomp(' ')
  end
end
```

something along these lines.

berziiii:
4 / 6. Nice work. Enumerables are there to make our lives easier and write less code!

PR #81, opened 2016-06-13
answers

mattb5:


lib/diagnostic.rb
```diff
@@ -5,36 +5,37 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    return list.max_by(&:length)
```
berziiii, line 8: You'll need the spaceship operator since it's a comparison

`list.max { |a , b| a.length <=> b.length }`

lib/diagnostic.rb
```diff
@@ -5,36 +5,37 @@ class Diagnostic
...
   end
 
+
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
```
berziiii, line 17: Correct.

lib/diagnostic.rb
```diff
@@ -5,36 +5,37 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.all? { |number| number.odd? }
```
berziiii, line 24: Correct. but shorthand, with enumerables, it

`numbers.all?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,37 @@ class Diagnostic
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |number| number.even? }
```
berziiii, line 31: Same here. 

`numbers.all?(&:even?)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,37 @@ class Diagnostic
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.each {|word| word.capitalize!}
```
berziiii, line 38: Again here.

`words.map(&:capitalize)` You need `.map` because you're changing the array

lib/diagnostic.rb
```diff
@@ -46,3 +47,5 @@ def read_file(filename)
     # Your code goes here
```
berziiii, line 47:
``` ruby
   file.each do |line|
    line.chomp(' ')
  end
end
```

Something along these lines.

berziiii:
3 / 6. Not bad. Remember enumerables make our lives easier!

PR #82, opened 2016-06-13
comfort: 3 clarity: 3

shireen-bean:
i used stack overflow

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -5,7 +5,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.select { |entry| entry.size.max }
   end
 
   # Question 2
@@ -13,28 +13,30 @@ class Diagnostic
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |number| number.odd? }
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |number| number.even? }
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.each do |word|
+      word.capitalize!
+    end
   end
 
   # Question 6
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -5,36 +5,38 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.select { |entry| entry.size.max }
```
berziiii, line 8: You need to use the spaceship operator since we're doing a comparison. 

lib/diagnostic.rb
```diff
@@ -5,36 +5,38 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
```
berziiii, line 16: Don't forget the &

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,38 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |number| number.odd? }
```
berziiii, line 23: Yes but the idea behind the enumerable methods is to reduce the amount of code we write!

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,38 @@ class Diagnostic
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |number| number.even? }
```
berziiii, line 30: Same here. Correct but short hand!

`numbers.all?(&:even?)`

lib/diagnostic.rb
```diff
@@ -5,36 +5,38 @@ class Diagnostic
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.each do |word|
+      word.capitalize!
```
berziiii, line 38: Looks liket his would work, but again, short hand

`words.map(&:capitalize)`

lib/diagnostic.rb
```diff
@@ -43,6 +45,9 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    File.open('filename').each do |line|
+      line.sub! newline, ' '
```
berziiii, line 49: You are missing a step. When you open the file, you a file to iterate through, you have to get down to a line

``` ruby
   file.each do |line|
    line.chomp(' ')
   end
end
```

berziiii:
4 / 6. Nice work. Remember, short hand with enumberables

PR #84, opened 2016-06-13
Responses to Ruby Enumerable Diagnostic

andreasantarl:
Comfort: 2
Clarity: 3

Not feeling great about RUby.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -5,6 +5,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
+    list.max_by{ &:length }
     # Your code goes here
   end
 
@@ -13,6 +14,7 @@ class Diagnostic
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
+    list.reduce (:+)
     # Your code goes here
   end
 
@@ -20,6 +22,7 @@ class Diagnostic
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
+    numbers.any?{ |number| number.odd? }
     # Your code goes here
   end
 
@@ -27,6 +30,7 @@ class Diagnostic
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
+    numbers.all? { |number| number.even?}
     # Your code goes here
   end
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -5,6 +5,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
+    list.max_by{ &:length }
```
berziiii, line 8: On the right path. Since it's a comparison, we need to use the shapeship operator.

lib/diagnostic.rb
```diff
@@ -13,27 +14,31 @@ def longest(list)
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
+    list.reduce (:+)
```
berziiii, line 17: Correct but don't forget the &

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -13,27 +14,31 @@ def longest(list)
...
     # Your code goes here
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
+    numbers.any?{ |number| number.odd? }
```
berziiii, line 25: Correct. but shorthand would be:

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -13,27 +14,31 @@ def longest(list)
...
     # Your code goes here
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
+    numbers.all? { |number| number.even?}
```
berziiii, line 33: Again correct but short hand would be:

`numbers.all?(&:even?)`

lib/diagnostic.rb
```diff
@@ -13,27 +14,31 @@ def longest(list)
...
     # Your code goes here
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
+    words.each { |word| word.capitalize }
```
berziiii, line 41: Again here. Short hand, which is what it's asking:

`words.map(&:capitalize)` You would use `.map` because you're changing the array itself.

lib/diagnostic.rb
```diff
@@ -44,5 +49,6 @@ def transform(words)
   # space and return the result.
   def read_file(filename)
     # Your code goes here
+    filename.foreach ('filename.csv') do |line| puts line += " "
```
berziiii, line 52: Almost

``` ruby
File.open('filename.csv') do |file|
   file.each do |line|
     line.chomp + " "
   end
end
```

berziiii:
4 / 6. The great thing about enumerable methods are that they help us write short hand code!

PR #86, opened 2016-06-13
Response

esthercordova:


lib/diagnostic.rb
```diff
@@ -6,35 +6,42 @@ class Diagnostic
   # a Fixnum.
   def longest(list)
     # Your code goes here
+    list.max
```
berziiii, line 9: Started off right!

What you have to do is to compare each element of the array with another. 

SO you have to iterate through and use the shaceship comparison to check the values of things

lib/diagnostic.rb
```diff
@@ -6,35 +6,42 @@ class Diagnostic
...
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
-  # Assume each entry implements a `+` method.
+  # Assume each entry implements  a `+` method.
   def sum(list)
     # Your code goes here
+    sum = 0
+    list.each.to_i { |i| sum += i }
```
berziiii, line 19: Yes but as the instructions say, and what we learned, you can call `list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -6,35 +6,42 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
     # Your code goes here
+   numbers.any? { |x| x.to_i % 2 != 0}
```
berziiii, line 27: Yes again, but short hand would be `numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -6,35 +6,42 @@ class Diagnostic
...
   end
 
-  # Question 4
+
+  # Question 4d
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
     # Your code goes here
+     numbers.all? { |x| x.to_i % 2 == 0 }
```
berziiii, line 36: Again yes. SHort hand: `numbers.all?(&:even?)`

lib/diagnostic.rb
```diff
@@ -6,35 +6,42 @@ class Diagnostic
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
     # Your code goes here
+    words.first.upcase
```
berziiii, line 44: Close `words.map(&:capitalize)`. Remember that `.map` re-writes each element of the array to what every you're doing to the element.

lib/diagnostic.rb
```diff
@@ -44,5 +51,12 @@ def transform(words)
   # space and return the result.
   def read_file(filename)
     # Your code goes here
+  File.open('filename') do |file|
+  str = ""
+  file.each do |line|
+  str.concat(line)
```
berziiii, line 57: Close. This way you'd have to `.chomp` the line and add ' ' . 

`line.chomp +  str` in this case.

berziiii:
3 / 6. You were close on all of the them. The methods are just used to shorten what we actually have to right but for 3 of the questions, you would have gotten the right answer with your tactic.

PR #88, opened 2016-08-05
David Powell, Diagnostic

dvdpwll:
I couldn't run rake test for some reason.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -5,7 +5,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max_by(&:length)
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -5,36 +5,38 @@ class Diagnostic
...
-    # Your code goes here
+    # numbers.all? { |num| num.even? }
+    numbers.all?(even?)
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
```
laurenfazah, line 37: `.each` won't return a new array. you'd want to use `.map`

laurenfazah:
5/6 good work

PR #93, opened 2016-08-05
Add answers to Ruby Enumerable Diagnostic

jlmitch1989:
Comfort: 2
Clarity: 2

I tried using notes, documentation, Google but I'm getting really confused
similarly to what confused me about javascript (not knowing and understanding
the structure).  I need some extra help.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -13,7 +13,7 @@ class Diagnostic
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    
   end
 
   # Question 3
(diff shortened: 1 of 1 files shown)


laurenfazah:
Glad to see you made an effort despite confusion. Refer to these answers below and let me know if any aren't making sense.

``` rb

class Diagnostic
  # Question 1
  # a Fixnum.
  def longest(list)
    # Your code goes here
  end

  # Question 2
  def sum(list)
    # Your code goes here
    list.reduce(&:+)
  end

  # Question 3
  def some_odd(numbers)
    # Your code goes here
    numbers.any?(&:odd?)
  end

  # Question 4
  # list of numbers are even
  def every_even(numbers)
    # Your code goes here
    numbers.all?(&:even?)
  end

  # Question 5
  def transform(words)
    # Your code goes here
    words.map(&:capitalize)
  end

  # Question 6
  def read_file(filename)
    # Your code goes here
    # File.open(filename).each_with_object('') do |line, memo|
    #   memo << line.chomp + ' '
    # end.strip
  end
end
```

PR #96, opened 2016-08-05
answered ?s

cdelsoin:


lib/diagnostic.rb
```diff
@@ -5,29 +5,29 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.sort { |a, b| b.length <=> a.length }
   end
```
laurenfazah, line 9: this would sort, but not return the longest element. 

``` rb
  def longest(list)
  end
```

lib/diagnostic.rb
```diff
@@ -5,29 +5,29 @@ class Diagnostic
...
+    list.reduce(:+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.odd?.reduce(:+)
   end
```
laurenfazah, line 25: This isn't going to do much other than return false (because of `numbers.odd?`)

``` rb
  def some_odd(numbers)
    numbers.any?(&:odd?)
  end
```

lib/diagnostic.rb
```diff
@@ -5,29 +5,29 @@ class Diagnostic
...
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+     numbers.even? ? true : false
   end
```
laurenfazah, line 31: Similar to Q3

``` rb
  def every_even(numbers)
    numbers.all?(&:even?)
  end
```

lib/diagnostic.rb
```diff
@@ -5,29 +5,29 @@ class Diagnostic
...
```
laurenfazah, line 32: Q5

``` rb
  def transform(words)
    words.map(&:capitalize)
  end
```

Q6

``` rb
  def read_file(filename)
    # File.open(filename).each_with_object('') do |line, memo|
    #   memo << line.chomp + ' '
    # end.strip
  end
```

laurenfazah:
good effort in the right direction...1/6

PR #122, opened 2016-11-14
respond to diagnostic

ioncrash:
now try bonus

comfort: 5
clarity: 5

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,13 +7,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    list.reduce { |prev, curr|
-      if prev.length > curr.length
-        return prev
-      else
-        return curr
-      end
-    }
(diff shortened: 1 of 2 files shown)


lib/diagnostic.rb
```diff
@@ -46,6 +46,12 @@ def transform(words)
...
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    result = ''
+    File.open(filename, "r") do |f|
+      f.each_line do |line|
+        result = result + line.chomp
+      end
+    end
+    result
```
neugierige, line 55: You're very close. Did you run the test on this? It looks like some of the returns are eliminated, but not replaced by a space.

neugierige:
5/5. Good work! I'd take another look at the bonus question.

lib/diagnostic.rb
neugierige: FYI, here's 1 solution:

```
  def read_file(filename)
    File.open(filename).reduce(String.new) { |a, e| a << e.chomp + ' ' }.strip
  end
```

PR #124, opened 2016-11-14
Responds to Ruby Enumerable Diagnostic

Naki21:


lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
-    # Your code goes here
+    numbers.all? { |e| e.even? }
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map {|e| e.capitalize}
```
neugierige, line 39: FYI, for questions 3, 4 and 5, you could actually use the `&:[method]` shorthand instead of writing the method inside the `{}`s

neugierige:
5/5. 🎉

PR #125, opened 2016-11-14
Attempted Enumerable Diagnostic on 11/14/16

Sphilippe7:
Did not do so well on this diagnostic.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max { a. <=> length }
   end
 
   # Question 2
@@ -15,21 +15,26 @@ class Diagnostic
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    odds = rng.select(&.odd?)
+
+  p odds
+
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    evens = rng.select(&.even?)
+
+  p evens
   end
 
   # Question 5


lib/diagnostic.rb
```diff
@@ -7,29 +7,34 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max { a. <=> length }
```
neugierige, line 10: Your're close. Here's 1 possible solution:

``` rb
def longest(list)
end
```

Notice that 1) length is a property of an object, and the 2) you need variables to capture those objects before you can get properties off of them, or operate with them.

lib/diagnostic.rb
```diff
@@ -7,29 +7,34 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+
```
neugierige, line 18: if you don't write down _anything_, we can't give you feedback

``` rb
  def sum(list)
    list.reduce(&:+)
  end
```

lib/diagnostic.rb
```diff
@@ -7,29 +7,34 @@ class Diagnostic
...
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    odds = rng.select(&.odd?)
+
+  p odds
+
   end
```
neugierige, line 29: a few things to consider whenever writing code:
1) are you utilizing the variable name of the argument inside your code at all? if not, you're doing something wrong.
2) look at all the words in your code. Which of them are variables? Have they been defined?
3) `p` is a method to print out something. Is that what the question is asking? Is that what you are intending to do?

Please read over the documentation for the `any?` method: http://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-any-3F

lib/diagnostic.rb
```diff
@@ -7,29 +7,34 @@ class Diagnostic
...
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    evens = rng.select(&.even?)
+
+  p evens
```
neugierige, line 37: Please read my comments for question 3, and look at the documentation for the `all?` method: http://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-all-3F

neugierige:
Hi Stanley, I know this was a hard one, but what you wrote in your responses revealed some gaps in your understanding of how to approach a coding program. Please read over my comments carefully and schedule a 1:1 to review this.

PR #126, opened 2016-11-14
Diagnostic2 Mon Nov 14

RpRoderick:
Answered questions, working through rake test now, ran out of time to
correct all fails

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max_by(&:length)
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max_by(&:length)
```
neugierige, line 10: nice!

lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
...
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
-  def some_odd(numbers)
-    # Your code goes here
-  end
 
+def some_odd(numbers)
+  put 'true' if numbers.any?(&:odd?)
+end
```
neugierige, line 27: You almost had it! Look again at what question asks. It says to `return true` if some condition is met, but your code says to `put 'true'`.

lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
...
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
+
   def every_even(numbers)
-    # Your code goes here
+    put 'true' if numbers.all?(&:even?)
   end
```
neugierige, line 34: same as above

lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
...
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.gsub(/\S+/, &:capitalize)
```
neugierige, line 40: The instructions tell us that `words` is a list of words, as the name suggests. You called `.gsub` on words, which is a method of String: https://ruby-doc.org/core-2.2.0/String.html#method-i-gsub

Here's 1 solution:

```
  def transform(words)
    words.map(&:capitalize)
  end
```

neugierige:
2/5. Ryan, you are totally on the right track, except you didn't read the directions closely enough!

PR #127, opened 2016-11-14
Commit Responses to Diagnostic

rickdean:
Diagnostic responses

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    longest.map(&:length)
   end
 
   # Question 2
@@ -15,28 +15,28 @@ class Diagnostic
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    sum.reduce(0, :+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    some_odd.any(&:odd?)
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    ever_even.all(& :even?)
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    transform.map(& :upcase)
   end
 
   # Question 6 (optional)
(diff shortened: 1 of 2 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    longest.map(&:length)
```
neugierige, line 10: Why would `.map` be the right choice here? the question says that the method should return a Fixnum. That's a single value. What does `.map` return? Take a look at the documentation: http://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-map

Here's 1 solution:

``` rb
  def longest(list)
  end
```

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    sum.reduce(0, :+)
```
neugierige, line 18: What does `sum.reduce` do? First of all, what is `sum`? It is a method that you have defined. So, in effect, you are calling your method from inside your method...

Secondly, notice that there is argument being passed into your function. Are you using that argument at all? 

Here's 1 solution:

``` rb
  def sum(list)
    list.reduce(:+)
  end
```

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    some_odd.any(&:odd?)
   end
```
neugierige, line 26: same issues here as above

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    ever_even.all(& :even?)
   end
```
neugierige, line 33: same

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    transform.map(& :upcase)
```
neugierige, line 39: same, and also, what is `:upcase`? Does it produce it result that you were looking for? Here's the documentation: https://ruby-doc.org/core-2.2.0/String.html#method-i-upcase

neugierige:
Rick, you were clearly able to correctly utilize the stuff we had done in class in this diagnostic. However, you made the SAME fundamental mistake in all of your questions. This is actually really important, and I hope this will be an opportunity for you to really internalize this.

PR #128, opened 2016-11-14
ruby enumerable diagnostic complete

akadane87:
Comfort: 2
Clarity: 2

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max_by(&:length)
   end
 
   # Question 2
@@ -15,28 +15,28 @@ class Diagnostic
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.inject(0, :+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.values_at(* numbers.each_index.select {|i| i.odd? })
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.values_at(* numbers.each_index.select {|i| i.even? })
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.chomp.sub(/./, &:upcase).to_a
   end
 
   # Question 6 (optional)


lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max_by(&:length)
```
neugierige, line 10: 👍

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.inject(0, :+)
   end
```
neugierige, line 19: another solution:

``` rb
  def sum(list)
    list.reduce(:+)
  end
```

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
-    # Your code goes here
+    numbers.values_at(* numbers.each_index.select {|i| i.odd? })
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.values_at(* numbers.each_index.select {|i| i.even? })
```
neugierige, line 32: Here's 1 solution:

``` rb
  def every_even(numbers)
    numbers.all?(&:even?)
  end
```

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
+    list.inject(0, :+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.values_at(* numbers.each_index.select {|i| i.odd? })
   end
```
neugierige, line 26: For Qs 3 and 4, I'm not sure where you got values_at from, but that's a method on Hashes. What is the datatype of numbers? It is a list.

Here's 1 solution:

``` rb
 def some_odd(numbers)
    numbers.any?(&:odd?)
  end
```

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
+    numbers.values_at(* numbers.each_index.select {|i| i.even? })
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.chomp.sub(/./, &:upcase).to_a
   end
```
neugierige, line 40: Here's the documentation for the `.chomp` method: https://ruby-doc.org/core-2.2.0/String.html#method-i-chomp

1st of all, notice that it is a method on String. What is the datatype of `words`? 2nd, look at the examples in the documentation and notice that in the 6th example, `"hello \n there".chomp` returns `"hello \n there"`. Is that what you were trying to achieve?

neugierige:
2/5. Hi Alex, I know this was a hard one. Please read over my comments and make sure you understand what I wrote. Remember, no one expects you to _memorize_ method syntax. You should _always_ refer to the documentation.

PR #129, opened 2016-11-14
Add response.

coffeeisamazing:
Add response.

Comfort: 4
Clarity: 4

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,9 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.inject do |memo, word|
+      memo.length > word.length ? memo : word
+    end
   end
 
   # Question 2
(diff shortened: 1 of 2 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +7,38 @@ class Diagnostic
...
-    # Your code goes here
+    numbers.all?(&:even?)
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map(&:upcase)
```
neugierige, line 41: what does `:upcase` do exact? Take a look at the documentation: https://ruby-doc.org/core-2.2.0/String.html#method-i-upcase

Would it transform 'foo' to 'Foo'?

lib/diagnostic.rb
```diff
@@ -46,6 +48,9 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    result = ''
+    File.open(filename).each do |line|
+      result += line
+    end
```
neugierige, line 54: You're on the right track! If you ever find yourself with extra time on your hands, I would encourage you to come back to this.

neugierige:
4/5. Good work! You made 1 small error on question 5. Take a look at my comments and let me know if you need another hint.

PR #130, opened 2016-11-14
Response to afternoon diagnostic

matthewgsommers:
Felt confident. Would like to check answers, but I knew what direction
to be moving in for each.

Clarity: 4
Comfort: 4

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+  list.sort_by(&:length)
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +7,39 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+  list.sort_by(&:length)
```
neugierige, line 10: The questions asks you to implement a method that returns "the longest entry." What does `sort_by` return? Here's the documentation: http://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-sort_by

Here's 1 solution:

``` rb
  def longest(list)
  end
```

lib/diagnostic.rb
```diff
@@ -7,36 +7,39 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.odd?
+    'true'
   end
```
neugierige, line 27: So `numbers` is a list. Is `.odd` a method we can call on lists? 

Take a look at this solution:

``` rb
  def some_odd(numbers)
    numbers.any?(&:odd?)
  end
```

Notice that the question asks if _some_ of the entries is odd, which is why we use `.any?`.

lib/diagnostic.rb
```diff
@@ -7,36 +7,39 @@ class Diagnostic
...
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    word = []
+    @words = words.normalize
+    word << words
   end
```
neugierige, line 43: what is `.normalize`? Is it a method you know that exists for the datatype of list, or a method that you wrote yourself? 

``` rb
  def transform(words)
    words.map(&:capitalize)
  end
```

neugierige:
2/5. Please review my comments. With Ruby, the benefit is that we have SO many methods available to us. However, we need to know exactly how to use them in order to achieve the desire result.

PR #131, opened 2016-11-14
response

kfili:


lib/diagnostic.rb
```diff
@@ -7,36 +7,39 @@ class Diagnostic
...
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    if numbers.some_odd == false
+       p true
+    end
   end
```
neugierige, line 36: 2 things: 1) what is `.some_odd`? does it exist? 2) look at what the question says: "returns true if..." What is your method doing? It's printing out true.

neugierige:
4/5. Good job! Take a look at my comments.

PR #132, opened 2016-11-14
Complete Ruby Enumerable Diagnostic

kmg1434:
respond to prompts in lib/diagnostic.rb

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,8 +6,14 @@ class Diagnostic
   # Write ruby code to return the longest entry in the list supplied.
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
+
   def longest(list)
-    # Your code goes here
+    list.each{ |prev, e|
+      if e.length > prev.length
+        max = e
+      end
+    }
+    max
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -6,37 +6,43 @@ class Diagnostic
...
+
   def longest(list)
-    # Your code goes here
+    list.each{ |prev, e|
+      if e.length > prev.length
+        max = e
+      end
+    }
+    max
   end
```
neugierige, line 17: Recall that `.each` is a method that handles individual elements in an Array. Here's the documentation: https://ruby-doc.org/core-2.2.0/Array.html#method-i-each

As such, you have 2 errors here: 1) your input, `list`, is not an array, 2) `.each` only works with 1 thing at a time

neugierige, line 17: Here's 1 solution: 

``` rb
  def longest(list)
  end
```

lib/diagnostic.rb
```diff
@@ -6,37 +6,43 @@ class Diagnostic
...
+    list.reduce(:+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any?{ |num| num.odd?}
   end
```
neugierige, line 32: FYI, for questions 3 and 4, you can also use the shorthand:

``` rb
def some_odd(numbers)
   numbers.any?(&:odd?)
end
```

neugierige:
4/5. Good work!

PR #134, opened 2016-11-14
Response

jbuzzanco:


lib/diagnostic.rb
```diff
@@ -7,6 +7,8 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
+    let longest_entry = list.max_by(&:length)
+
```
neugierige, line 11: Is `let` a thing in Ruby? 🤔 

you're very close...

lib/diagnostic.rb
```diff
@@ -15,28 +17,31 @@ def longest(list)
...
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    if evens = numbers.select(&:evens?)
+      true
+    end
   end
```
neugierige, line 38: same problem as above, but you do have the right number of `end`s

lib/diagnostic.rb
```diff
@@ -15,28 +17,31 @@ def longest(list)
...
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
-  end
+    if odds = numbers.select(&:odd?)
+      true
+    end
```
neugierige, line 30: 1) I notice that you have a `def` and an `if` statement, but only 1 `end`.
2) what is line 27 doing? what is `odds`? and how can the if-statement evaluate an assignment expression? (recall that `=` is assignment, and `==` is checking equivalence)

lib/diagnostic.rb
```diff
@@ -15,28 +17,31 @@ def longest(list)
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    capitalized = words.upcase
   end
```
neugierige, line 45: Besides the fact that you are again assigning value to a variable, what exactly is `.upcase`? Did you look at the documentation to see if it gives you what you are going for?

neugierige:
1/5. Hi Joel, please read over my comments and make sure you understand them. With Ruby, it's great that we have so many methods we can use, but it's important to always check the documentation to see how it is correctly implemented.

PR #136, opened 2016-11-14
Submit diagnostic

akradin:
4 and 4
Use stack overflow and Ruby Docs

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.split(' ').max_by(&:length)
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.split(' ').max_by(&:length)
   end
```
neugierige, line 11: Very close. What does `.split` work with? is list the appropriate datatype?

What does `list.split(' ')` give you? is that the appropriate datatype on which to use `max_by`? take a look at the documentation for `.max_by`: http://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-max_by

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
+    numbers.all? { |i| i.even? }
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.split(' ').map { |i| i.capitalize }
   end
```
neugierige, line 40: same error here as in question 1...

neugierige:
3/5. Good work! Make sure you understand the error you made, since you did it twice...

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
+    list.reduce(:+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |i| i.odd? }
   end
```
neugierige, line 26: FYI, this can also be written with the shorthand:

``` rb
def some_odd(numbers)
  numbers.any?(&:odd?)
end
```

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |i| i.even? }
   end
```
neugierige, line 33: shorthand version also works for this one. can you see how?

PR #137, opened 2016-11-14
Complete diagnostic on Ruby Enums

sujkid:
Done.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,10 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    longest = list.inject do |memo, word|
+    memo.length > word.length ? memo : word
+    end
+    longest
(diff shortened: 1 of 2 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +7,39 @@ class Diagnostic
...
+    numbers.all? { |num| num.even? }
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map { |word| word.upcase! }
   end
```
neugierige, line 43: the question asks for words to capitalized, as in from 'foo' to 'Foo'. Does `.upcase` do that? Take a look at the documentation: https://ruby-doc.org/core-2.3.0/String.html#method-i-upcase

neugierige:
4/5. Good work! 1 minor error on question 5. Take a stab at question 6 when you have time!

PR #138, opened 2016-11-14
Complete ruby-enumerable diagnostic

kiraamaa:
Use Google and class resources to complete
Attempt final question

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max_by(&:length)
   end
 
   # Question 2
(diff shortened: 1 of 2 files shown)


kiraamaa:
Comfort: 4
Clarity: 4

lib/diagnostic.rb
```diff
@@ -46,6 +46,13 @@ def transform(words)
...
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    fh = open filename
+    # need to read each line
+    fh.each do |line|
+      # need to replace newline character with a space
+      line.replace(' ')
+      # need to concatenate and return
+      line.join
```
neugierige, line 55: `.join` is a method on Arrays. Is line an array? 🤔 
https://ruby-doc.org/core-2.2.0/Array.html#method-i-join

neugierige:
5/5. Good work! You're pretty close with question 6. I'd take another look at it if I were you...

PR #139, opened 2016-11-14
ruby-enumerable-diagnostic response

treep78:
Comfort: 4
Clarity: 4

I still have issues with num 5.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -2,12 +2,13 @@
 
 # Ruby enumerable diagnostic
 class Diagnostic
+  include Enumerable
   # Question 1
   # Write ruby code to return the longest entry in the list supplied.
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max_by(&:length)
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -2,41 +2,42 @@
...
+    numbers.all? {|num| num%2 == 0}
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    p words.entries {|word| "#{word[0].upcase}#{word[1, word.length]}"}
   end
```
neugierige, line 41: a few things to think about:
1) what was your expectation for the behavior of `.entries`? take a look at the documentation again: http://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-entries
2) in the block of code that comes after, what is `word`? what did you want it to be, and what is it in the code you wrote? 
3) assuming that you are getting the characters of a word by putting the index/range in the `[]`s, what is the value of `word.length`? If you are starting at index 1, what would be the appropriate length from there to the end of the string?

neugierige, line 41: Here's 1 solution:

``` rb
  def transform(words)
    words.map(&:capitalize)
  end
```

lib/diagnostic.rb
```diff
@@ -2,41 +2,42 @@
...
+    numbers.any? {|num| num%2 != 0}
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? {|num| num%2 == 0}
   end
```
neugierige, line 34: Both 3 & 4 work perfectly fine, but you are not taking advantage of the methods Ruby provides for us. Consider this solution:

``` rb
  def every_even(numbers)
    numbers.all?(&:even?)
  end
```

https://ruby-doc.org/core-2.2.2/Integer.html#method-i-even-3F

neugierige:
4/5. Please look over my comments, and see if you can figure out how to fix your response to question 5. You're very close!

PR #140, opened 2016-11-14
Answer diagnostic questions.

jdaly04:
Add responses to ruby-enumerable-diagnostic.
Comfort: 2
Clarity: 3

Used google and used training repos from class.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max_by(&:length)
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
+    numbers.all? { |number| number%2 == 0 }
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+      words.split.map(&:capitalize)  # haven't finished! not sure if that's right!
   end
```
neugierige, line 40: you're very close! recall that `.split` is a method on Arrays: https://ruby-doc.org/core-2.3.0/String.html#method-i-split

`words` is a list, so `.split` wouldn't work. See what happens if you just remove that 1 term...

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
+    numbers.any? { |number| number%2 != 0 }
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |number| number%2 == 0 }
   end
```
neugierige, line 33: bother 3 & 4 are fine as is, but here's a shorthand way to doing the same thing:

``` rb
  def every_even(numbers)
    numbers.all?(&:even?)
  end
```

neugierige:
4/5. Excellent work, Julia!

PR #141, opened 2016-11-14
Response to diagnostic

ielole:
Attempt

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.select { |num| }
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +7,38 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.select { |num| }
   end
```
BenGitsCode, line 11:
``` ruby
```

neugierige, line 11: here is the documentation for `.select`: https://ruby-doc.org/core-2.2.0/Array.html#method-i-select

Notice that `.select` returns an array that match the condition specified in the block. Is that what we are trying to accomplish here? The question said to "return the longest entry."

lib/diagnostic.rb
```diff
@@ -7,36 +7,38 @@ class Diagnostic
...
+    end
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |n| n.odd == true }
   end
```
neugierige, line 28: `.odd` is not a method that exists in Ruby. I think what you intended was this: https://ruby-doc.org/core-2.2.2/Integer.html#method-i-odd-3F

Notice that this method immediately returns a `true` or `false`. While `true == true` does evaluate to `true`, it's a bit redundant. Consider this possible solution:

``` rb
  def some_odd(numbers)
    numbers.any?(&:odd?)
  end
```

lib/diagnostic.rb
```diff
@@ -7,36 +7,38 @@ class Diagnostic
...
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |n| n.even == true }
   end
```
neugierige, line 35: same as above, here's the documentation for `.even?`: https://ruby-doc.org/core-2.2.2/Integer.html#method-i-even-3F

lib/diagnostic.rb
```diff
@@ -7,36 +7,38 @@ class Diagnostic
...
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.
   end
```
neugierige, line 42: word.

here's 1 solution:

``` rb
  def transform(words)
    words.map(&:capitalize)
  end
```

neugierige:
1/5. Ariel, you were _very_ close on questions 3 and 4, but you off by literally 1 character! Please look over my comments and let me know if you have questions.

PR #151, opened 2017-04-14
Commit for ruby enumerable diagnostic

deeforte:
commit for April 14th diagnostic on ruby enumerables
comfort: 3
clarity: 3

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,8 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.inject do |list, word|
+      list.length > word.length ? list : word
   end
 
   # Question 2
@@ -15,28 +16,28 @@ class Diagnostic
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list_sum = list.reduce(:+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |num| num.odd? }
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |num| num.even? }
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.capitalize!
   end
 
   # Question 6 (optional)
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.inject do |list, word|
+      list.length > word.length ? list : word
```
jordanallain, line 11: Here's an easy Ruby way to do it:

`list.max_by(&:length)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list_sum = list.reduce(:+)
```
jordanallain, line 19: This works but using the `&` is safer!

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |num| num.odd? }
```
jordanallain, line 26: This works too but here's the quick Ruby way:

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |num| num.even? }
```
jordanallain, line 33: Cool, but again here's the quicker Ruby-er way:

`numbers.all?(&:even?)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.capitalize!
```
jordanallain, line 40: This wouldn't capitalize every word, this would work:

`words.map(&:capitalize)`

Using map would call it on every word.

lib/diagnostic.rb
```diff
@@ -46,6 +47,10 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    File.open(filename) do |file|
+      file.each do |line, i|
+      all_lines = all_lines + line + " "
+    end
+   end
```
jordanallain, line 54: You were on the right track here, this would work:

```ruby
    a << e.chomp + '' }.strip
```

jordanallain:
3/6

Good job Dee Dee, very close on a couple of them.

🇯🇴 🇯🇴 🇯🇴

PR #152, opened 2017-04-14
Submitting response

sjruby:
Comfort 3/5
Clarity 3/5

I feel good about Ruby, there were some funcitons that I couldnt get to
work as the short hand {: even?} when the linter was yelling at me so I
left the sugarless and they seemed to work.  Couldnt get the rake test ot
work so I created my own test variable and messed with them in Pry. Seemed
to work.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-
+list = 1..10
 # Ruby enumerable diagnostic
 class Diagnostic
   # Question 1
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -1,44 +1,53 @@
...
+list = 1..10
 # Ruby enumerable diagnostic
 class Diagnostic
   # Question 1
   # Write ruby code to return the longest entry in the list supplied.
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max
```
jordanallain, line 10: Using `.max` we would usually use a block so it knows how to compare each element:

`list.max { |a, b}| a.length <=> b.length }`

lib/diagnostic.rb
```diff
@@ -1,44 +1,53 @@
...
   end
-
+  longest(list)
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
```
jordanallain, line 18: This would work but its safer to use the `&` symbol:

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -1,44 +1,53 @@
...
   end
+  sum(list)
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
-  def some_odd(numbers)
-    # Your code goes here
+  def some_odd(list)
+    list.any? { :odd? }
```
jordanallain, line 26: Looks like you gave it a block, should have been parens:

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -1,44 +1,53 @@
...
   end
+  some_odd(list)
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    # numbers.select { :even? } NOT SURE WHY THIS ISN:T WORKING BUT WHEN I TEST IT BOMBS
+    numbers.select { |num| num.even? }
```
jordanallain, line 35: Again you are using braces and making a block but you should just use parens:

`numbers.all?(&:even?)`

lib/diagnostic.rb
```diff
@@ -1,44 +1,53 @@
...
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
+  test_list = ['foo','foo','fooooooo']
+
   def transform(words)
-    # Your code goes here
+    words.each { |word| word.capitalize! }
+    # words.each { :capitalize! }
```
jordanallain, line 46: This works but here's a quicker way:

`words.map(&:capitalize)`

jordanallain:
2/5

You missed a couple just from using braces instead of parens. Good job otherwise!

🇯🇴 🇯🇴

PR #154, opened 2017-04-14
Completes diagnostic.rb

buffalo0707:
Comfort: 5
Clarity: 5

Feeling pretty good about this stuff, but didn't quite get the last
optional question. Rake test showed the method was returning an error
which I wasn't able to debug in time. Oh well.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max { |a, b| a.length <=> b.length}
   end
 
   # Question 2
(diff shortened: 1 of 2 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max { |a, b| a.length <=> b.length}
```
jordanallain, line 10: This works, here's a quicker way we can do it:

`list.max_by(&:length)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce { |sum, e| sum + e }
```
jordanallain, line 18: Cool, and a quicker way:

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -46,6 +46,13 @@ def transform(words)
...
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    result = ''
+    File.open(filename) do |file|
+      file.each do |line|
+        result += line.strip + ' '
+      end
+    end
+      puts result
```
jordanallain, line 55: This didn't pass the tests but it looks like you were on the right track. Check out the solution code if you need a nudge in the right direction.

jordanallain:
5/6

Great work Alan! Let me know if you need to go over the last answer.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #155, opened 2017-04-14
Add responses to ruby enumerable diag

lisamsmith100:
FOF: 3, 3, used class notes and repo documentation, + ruby document

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,29 +7,52 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
-  end
+
+  include Comparable
+
+    def <=>(list)
+      [RANKS.index(rank), SUITS.index(suit)] <=>
+        [RANKS.index(other.rank), SUITS.index(other.suit)]
+
+
+  #   include Enumerable
+  #
+  #   def each
+  #     @list = list
+  #     current_list_item = 0
+  #
+  #     while current_list_item < list.length
+  #
+  #       if current_list_item[i] < current_list_item[i+1]
+  #         yield current_list_item
+  #         current_list_item += 1
+  #       end
+  #     end
+  # end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -7,29 +7,52 @@ class Diagnostic
...
+  #     current_list_item = 0
+  #
+  #     while current_list_item < list.length
+  #
+  #       if current_list_item[i] < current_list_item[i+1]
+  #         yield current_list_item
+  #         current_list_item += 1
+  #       end
+  #     end
+  # end
```
jordanallain, line 31: Whoa got a lot going on for this first one.  You probably overthought it, here's a quick way we can do it:

`list.max_by(&:length)`

lib/diagnostic.rb
```diff
@@ -7,29 +7,52 @@ class Diagnostic
...
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
```
jordanallain, line 38: This does work but using the ampersand is safer:

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -7,29 +7,52 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? {|num| num.even}
```
jordanallain, line 45: Close, here:

`numbers.any?(&odd?)`

lib/diagnostic.rb
```diff
@@ -7,29 +7,52 @@ class Diagnostic
...
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers_set
+    numbers_set.each { |numbers_set, i|
+      numbers % 2 == 0
+    }
```
jordanallain, line 55: Here's the easy way to do it:

`numbers.all?(&:even?)`

lib/diagnostic.rb
```diff
@@ -45,7 +68,17 @@ def transform(words)
...
+    input_file = ARGV[0]
+
+    File.open(input_file) do |file|
+      file.each_with_index do |line, i|
+        "#{i}: #{line}"
+
+      end
+    end
+      end
+    end
```
jordanallain, line 84: Cool looks like you were on the right track!

```ruby
    a << e.chomp + ' ' }.strip
```

jordanallain:
1/6

You got really close on a few of these Lisa! Let us know if you need a hand with any.

🇯🇴

PR #156, opened 2017-04-14
study for 4-14

awoodrum87:
2/5 for comfort and clarity

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 # Ruby enumerable diagnostic
+list = [0,1,1,2,3,5,8,13,21,34]
 class Diagnostic
   # Question 1
   # Write ruby code to return the longest entry in the list supplied.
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -15,21 +16,21 @@ def longest(list)
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
```
jordanallain, line 19: This works but using the ampersand is safer:

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -15,21 +16,21 @@ def longest(list)
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? {|n| n if n % 2 != 0}
```
jordanallain, line 26: Here's an even quicker way to do it:

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -15,21 +16,21 @@ def longest(list)
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? {|n| n if n % 2 == 0}
```
jordanallain, line 33: Cool and another more Ruby way to do it:

`numbers.all?(&:even?)`

jordanallain:
3/6

Good job Angie! Check out the solution code to the ones you were stuck on if you want to see the answers.

🇯🇴 🇯🇴 🇯🇴

PR #160, opened 2017-04-14
my response for the ruby enumerable diagnostic

gastevend:


lib/diagnostic.rb
```diff
@@ -7,36 +7,40 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.reduce { |long, word|
+      if word.length >= long.length
+        long = word
+      end
+    }
```
jordanallain, line 14: Got a little verbose here, there's usually always a quick way to do things in Ruby like this:

`list.max_by(&:length)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,40 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
```
jordanallain, line 22: Good, its best to use the ampersand as well to be safe:

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,40 @@ class Diagnostic
...
-    # Your code goes here
+    numbers.any?(&:odd?)
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.any?(&:even?)
```
jordanallain, line 36: Really close but you should have used `all?` instead of `any?`

lib/diagnostic.rb
```diff
@@ -7,36 +7,40 @@ class Diagnostic
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map(&:capitalized!)
```
jordanallain, line 43: Super close again! Leave the `d` off of capitalize though and we don't need the mutator.

lib/diagnostic.rb
```diff
@@ -46,6 +50,7 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    filename.each_with_index do |line, index|
+    p line + ''
+    end
```
jordanallain, line 55: You were on the right track here:

```ruby
    a << e.chomp + ' ' }.strip
```

jordanallain:
2/6

You got caught up on a few of these but you were on the right track with most of them! 

🇯🇴 🇯🇴

PR #161, opened 2017-04-14
Response to Ruby Enumerable Diagnostic

rrdaniels85:
Comfort 4/5
Clarity 4/5

used notes from class and ruby documentation on methods from the
ruby site

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max
```
jordanallain, line 10: Close but we should pass `max` a block so it knows how to compare the elements:

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce { |sum, n| sum + n }
```
jordanallain, line 18: Good, here's an even quicker way to do it:

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any? { |num| num % 2 != 0 }
```
jordanallain, line 25: Good! Here's another way:

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.all? { |num| num % 2 == 0 }
```
jordanallain, line 32: Great, and again another quick way we can do it:

`numbers.all?(&:even?)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,36 @@ class Diagnostic
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    array.each {|i| i.capitalize}
```
jordanallain, line 39: Close, here:

`words.map(&:capitalize)`

jordanallain:
3/6

Good job Rachel, you missed a couple but you were on the right track.

🇯🇴 🇯🇴 🇯🇴

PR #162, opened 2017-04-14
Ruby enumerable diagnostic.

glasneed:
Fist of Five on Ruby.  3
NEed long weekend to review material from this week.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,37 +7,55 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    the_length
+    the_longest_string
+    the_longest_string = list.max_by(&:length)
+    the_length = the_longest_string.length # find length of longest string returned
+    return the_length
   end
+#Source: Slackoverflow. Tested in pry
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.inject(0){|sum,x| sum + x }
   end
+
+  #Source: Slackoverflow. Tested in pry
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
+  # I assumed list was size 10.
   def some_odd(numbers)
-    # Your code goes here
+    x = 1
+    returnValue = false
+    loop do
+      x += 1
+      next unless ( x % 2 ) != 0 # find the odd number
+      returnValue = true
+      break if x >= 10
+    end
+    return returnValue
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -7,37 +7,55 @@ class Diagnostic
...
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    the_length
+    the_longest_string
+    the_longest_string = list.max_by(&:length)
+    the_length = the_longest_string.length # find length of longest string returned
+    return the_length
   end
+#Source: Slackoverflow. Tested in pry
```
jordanallain, line 16: Hey Conor you wrote a lot more code than we need to do this in Ruby. Here's a quick way we can do it:

`list.max_by(&:length)`

Also remember that in Ruby it automatically returns the last line of the method so you don't need to use the `return` keyword.

lib/diagnostic.rb
```diff
@@ -7,37 +7,55 @@ class Diagnostic
...
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.inject(0){|sum,x| sum + x }
```
jordanallain, line 23: Here's a really quick way to do this in Ruby:

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -7,37 +7,55 @@ class Diagnostic
...
-    # Your code goes here
+    x = 1
+    returnValue = false
+    loop do
+      x += 1
+      next unless ( x % 2 ) != 0 # find the odd number
+      returnValue = true
+      break if x >= 10
+    end
+    return returnValue
```
jordanallain, line 41: Remember here we don't need the keyword `return`. Here's the quick Ruby way to do this problem:

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -7,37 +7,55 @@ class Diagnostic
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    numbers.each { |value| puts value if value.to_f%2==0}
```
jordanallain, line 48: Close here, this is better:

`numbers.all?(&:even?)`

lib/diagnostic.rb
```diff
@@ -7,37 +7,55 @@ class Diagnostic
...
   end
+  #Source: Google. Tested in pry
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map(&:upcase)
```
jordanallain, line 56: This was on the right track, here's a way to do it:

`words.map(&:capitalize)`

jordanallain:
0/6

Looks like you got a little mixed up with Ruby and JavaScript so let Chris know if you want to go over any of these.

PR #164, opened 2017-04-14
Add responses to diagnostic

jim-moody:
Clarity: 4
Comfort: 4

I used the Ruby documentation on String, Enumerator and Arrays to figure out how to complete the questions.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -1,4 +1,3 @@
-# frozen_string_literal: true
 
 # Ruby enumerable diagnostic
 class Diagnostic
@@ -7,7 +6,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.sort { |a, b| b.length <=> a.length }[0]
   end
 
   # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +6,36 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.sort { |a, b| b.length <=> a.length }[0]
```
jordanallain, line 9: This is super clever! Here's a quicker way we can do it though:

`list.max_by(&:length)`

lib/diagnostic.rb
```diff
@@ -7,36 +6,36 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce { |sum, number| sum + number }
```
jordanallain, line 17: Nice, here's another way:

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -45,7 +44,14 @@ def transform(words)
...
   # replacing the trailing newline character with a
   # space and return the result.
+
   def read_file(filename)
-    # Your code goes here
+    new_line = ''
+    File.open(filename) do |file|
+      file.each_line { |line| new_line.concat(line.split("\n")[0] + " ") }
+    end
+   new_line
```
jordanallain, line 53: Looks like you were on the right track here:

```ruby
    a << e.chomp + ' ' }.strip
```

jordanallain:
5/6

Great work as usual Jim! Showed you some quicker Ruby-er ways to do them.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #165, opened 2017-04-14
response

MattKwonDo:
3 3

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,7 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.max_by(&:length)
   end
 
   # Question 2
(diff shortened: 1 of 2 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +7,38 @@ class Diagnostic
...
+    list.max_by(&:length)
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.inject(0){ |sum, x| sum + x }
```
jordanallain, line 18: Close, here's a good way to do it:

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,38 @@ class Diagnostic
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    if numbers.any? { |num| num.odd? }
+      numbers.inject(0){ |sum, x| sum + x }
+    end
```
jordanallain, line 27: Here's an easy way to do this one:

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -46,6 +48,6 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    ('./data/cusa-preamble').head
```
jordanallain, line 51: Looks like you ran out of time, here's a way to do this:

```ruby
    a << e.chomp + ' ' }.strip
```

jordanallain:
3/6

Good job with these Matt, you were really close on a couple of them. If you need to go over any of the answers let us know!

🇯🇴 🇯🇴 🇯🇴

PR #166, opened 2017-04-14
Ruby enum diagnostic

antleo1995:
Answered the questions. Got all but the optional one.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,8 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.inject do | memo, word| memo.length > word.length ? memo : word
+    end
   end
 
   # Question 2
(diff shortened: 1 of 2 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.inject do | memo, word| memo.length > word.length ? memo : word
+    end
```
jordanallain, line 11: Good work Tony! Here's a quicker way to do it:

`list.max_by(&:length)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
```
jordanallain, line 19: Good here, its safer to use the ampersand though:

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
...
-    # Your code goes here
+    numbers.all?(&:even?)
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.each(&:capitalize!)
```
jordanallain, line 40: Good! This mutates the original array which actually isn't necessary. We can just return the result of calling the method but either way this works!

lib/diagnostic.rb
```diff
@@ -46,6 +47,6 @@ def transform(words)
   # replacing the trailing newline character with a
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    filename.open { |input, string| input.each + string }
```
jordanallain, line 50: Looks like you were on the right track here, this is one way to do it:

```ruby
    a << e.chomp + ' ' }.strip
```

jordanallain:
5/6

Awesome Tony! Almost got all of them.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #169, opened 2017-04-14
Response

donpowers:


lib/diagnostic.rb
```diff
@@ -7,36 +7,61 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    longest
+    list.each do |entry|
+      if longest.length < entry.length
+        longest = entry
+    end
+    longest
```
jordanallain, line 15: Here's a nice quick way we can do this in Ruby:

`list.max_by(&:longest)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,61 @@ class Diagnostic
...
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    sum = 0
+    list.each do |entry|
+      sum += entry
+    end
+    sum
```
jordanallain, line 27: Here's another quick way to do this one:

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,61 @@ class Diagnostic
...
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    nubmers.each do |entry|
+      if (entry.odd?)
+        return true
+      end
+    end
+    return false
```
jordanallain, line 39: Remember we don't need to use the `return` keyword in Ruby, the last line of the if statement or function will be returned for us. Here's how we can do this one:

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,61 @@ class Diagnostic
...
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+      numbers.each do |entry|
+        if (entry.odd?)
+          return false
+        end
+      end
+      return true
```
jordanallain, line 51: Here again we wouldn't need to say `return`. With Ruby there is often a built in method we can use to our advantage:

`numbers.all?(&:even?)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,61 @@ class Diagnostic
...
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    capWords
+    words.each do |entry i|
+      capwords[i] = entry.capitalize
+      end
+    end
+    return capWords
```
jordanallain, line 64: This one would be easier if we used an array method:

`words.map(&:capitalize)`

jordanallain:
0/6

Don it looks like you were overthinking these, or maybe trying to do them how you would in JavaScript. With Ruby there's a lot of built in methods we can take advantage of to do things really quickly. If any of these are still confusing we can go over them in a 1on1.

PR #170, opened 2017-04-14
Finished Ruby Enumerables Diagnostic

jscohen:
Clarity: 5
Comfort: 4

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -7,7 +7,8 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.map! {|item| item.length}
+    list.max
   end
 
   # Question 2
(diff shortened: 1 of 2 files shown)


lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    list.map! {|item| item.length}
+    list.max
```
jordanallain, line 11: Whoops, you are returning only the length instead of the element. Here's a quick way to do it:

`list.max_by(&:length)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list(:+)
```
jordanallain, line 19: Close, here:

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -7,36 +7,37 @@ class Diagnostic
...
-    # Your code goes here
+    numbers.all?(&:even?)
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.map!(&:capitlize!)
```
jordanallain, line 40: Looks like you spelled capitalize wrong, also you don't need the `!` because we don't want to overwrite the original array:

`words.map(&:capitalize)`

lib/diagnostic.rb
```diff
@@ -46,6 +47,12 @@ def transform(words)
...
   # space and return the result.
   def read_file(filename)
-    # Your code goes here
+    str = ''
+    File.open(filename) do |file|
+      file.each_with_index do |f|
+        str += f.sub!("\n", " ")
+      end
+    end
+    str
```
jordanallain, line 56: Here's one way to accomplish this:

```ruby
    a << e.chomp + ' ' }.strip
```

jordanallain:
3/6

Good work Jon, you were on the right track on the ones you missed.

🇯🇴 🇯🇴 🇯🇴

PR #171, opened 2017-04-14
Amanda's answers

aeabbott:
Needed a lot more time to finish this and check answers but I think I would
have gotten there eventually.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,37 +6,36 @@ class Diagnostic
   # Write ruby code to return the longest entry in the list supplied.
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
-  def longest(list)
-    # Your code goes here
-  end
+
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
+
   def some_odd(numbers)
-    # Your code goes here
+    numbers.any?{|number| number.odd?}
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    number.all? {|number| number.odd?}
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -6,37 +6,36 @@ class Diagnostic
   # Write ruby code to return the longest entry in the list supplied.
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
-  def longest(list)
-    # Your code goes here
-  end
+
```
jordanallain, line 9: Here's a quick way to do this one:

`list.max_by(&:length)`

lib/diagnostic.rb
```diff
@@ -6,37 +6,36 @@ class Diagnostic
...
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    list.reduce(:+)
```
jordanallain, line 16: This works but safer to use the ampersand as well:

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -6,37 +6,36 @@ class Diagnostic
...
-    # Your code goes here
+    numbers.any?{|number| number.odd?}
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    number.all? {|number| number.odd?}
```
jordanallain, line 31: I think this one works unless it gets the number 0 passed to it for some reason. Here's another way we can do it:

`numbers.all?(&:even?)`

lib/diagnostic.rb
```diff
@@ -6,37 +6,36 @@ class Diagnostic
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+    words.split.map(&:capitalize)
```
jordanallain, line 38: Super close here, just remove the `.split`:

`words.map(&:capitalize)`

jordanallain:
2/6

You were super close on a couple of the ones you missed, good job!

🇯🇴 🇯🇴

PR #173, opened 2017-04-14
Add response

khan-ngo:


lib/diagnostic.rb
```diff
@@ -2,41 +2,46 @@
...
 # Ruby enumerable diagnostic
 class Diagnostic
+
   # Question 1
   # Write ruby code to return the longest entry in the list supplied.
   # Assume each entry in the list implements a length method that returns
   # a Fixnum.
   def longest(list)
-    # Your code goes here
+    array.select{|char_num| char_num.size.max}
```
jordanallain, line 11: Here's a quick way to accomplish this:

`list.max_by(&:length)`

lib/diagnostic.rb
```diff
@@ -2,41 +2,46 @@
...
   end
 
   # Question 2
   # Write ruby code to return the sum of entries in the
   # list supplied (i.e. apply the `+` operator).
   # Assume each entry implements a `+` method.
   def sum(list)
-    # Your code goes here
+    array.each(:+)
```
jordanallain, line 19: `list` is the parameter so that is what you should be working with in the method:

`list.reduce(&:+)`

lib/diagnostic.rb
```diff
@@ -2,41 +2,46 @@
...
   end
 
   # Question 3
   # Write ruby code that returns true if some of the entries
   # in the list of numbers are odd
   def some_odd(numbers)
-    # Your code goes here
+    array.odd?
```
jordanallain, line 26: Again `numbers` is the parameter here:

`numbers.any?(&:odd?)`

lib/diagnostic.rb
```diff
@@ -2,41 +2,46 @@
...
   end
 
   # Question 4
   # Write ruby code that returns true if every entry in the
   # list of numbers are even
   def every_even(numbers)
-    # Your code goes here
+    # aray.select { |num| num.even?}
+
+    aray.select(&:even?)
```
jordanallain, line 35: Here's one way to do this one:

`numbers.all?(&:even?)`

lib/diagnostic.rb
```diff
@@ -2,41 +2,46 @@
...
   end
 
   # Question 5
   # Write ruby code that returns an array with all of the
   # entries from the list of words capitalized (i.e. 'foo' becomes 'Foo').
   def transform(words)
-    # Your code goes here
+
+    aray.select(&:upcase?)
+
```
jordanallain, line 44: We don't want to check for uppercase here we want to capitalize each word:

`words.map(&:capitalize)`

jordanallain:
0/6

Looks like you struggled with this one Khanh. Make sure you reach out to Chris and set up a 1on1 before the next project to review stuff that is still unclear to you.

